PR #115, opened 2015-03-17
Support for recent (4.1.0.x) V8 version

schwebke:
Recent V8 versions changed the API quite a bit. See the V8's project change notes
https://docs.google.com/document/d/1g8JFi8T_oAE_7uAri7Njtig7fKaPDfotU6huOa1alds/
or this blog post for a summary
http://strongloop.com/strongblog/node-js-v0-12-c-apis-breaking/

With the changes from this pull request plv8 builds with the 4.1.0.x V8 version. This version currently ships with chromium stable:
http://www.chromium.org/developers/calendar
Depending on the minor 4.1.0.x version the actually supported gcc versions vary. The compile flags try to compensate for this. gcc 4.8.2 on Ubuntu 14.04 is working fine here. In case of issues building V8 on older systems try falling back to an other V8 from 4.1 branch, e.g. 4.1.0.14.

### DIFF
--- a/Makefile.v8
+++ b/Makefile.v8
@@ -7,11 +7,14 @@
 # structure in v8 which may be different from version to another, but user
 # can specify the v8 version by AUTOV8_VERSION, too.
 #-----------------------------------------------------------------------------#
-AUTOV8_VERSION = 3.14.5
-AUTOV8_DIR = build/v8-$(AUTOV8_VERSION)
-AUTOV8_OUT = build/v8-$(AUTOV8_VERSION)/out/native
+AUTOV8_VERSION = 4.3.66
+AUTOV8_DIR = build/v8-git-mirror-$(AUTOV8_VERSION)
+AUTOV8_OUT = build/v8-git-mirror-$(AUTOV8_VERSION)/out/native
+AUTOV8_DEPOT_TOOLS = build/depot_tools
 AUTOV8_LIB = $(AUTOV8_OUT)/libv8_snapshot.a
-AUTOV8_STATIC_LIBS = libv8_base.a libv8_snapshot.a
+AUTOV8_STATIC_LIBS = libv8_base.a libv8_snapshot.a libv8_libplatform.a libv8_libbase.a libicui18n.a libicuuc.a libicudata.a
+
+export PATH := $(abspath $(AUTOV8_DEPOT_TOOLS)):$(PATH)
 
 SHLIB_LINK += $(addprefix $(AUTOV8_OUT)/, $(AUTOV8_STATIC_LIBS))
 
@@ -20,23 +23,40 @@ all: $(AUTOV8_LIB)
 # For some reason, this solves parallel make dependency.
 plv8_config.h: $(AUTOV8_LIB)
 
-$(AUTOV8_DIR):
+$(AUTOV8_DEPOT_TOOLS):
 	mkdir -p build
-	curl -L 'https://github.com/v8/v8/archive/$(AUTOV8_VERSION).tar.gz' \
-		| tar zxf - -C build
+	cd build; git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+
+$(AUTOV8_DIR): $(AUTOV8_DEPOT_TOOLS)
+	cd build; fetch v8; cd v8; git checkout $(AUTOV8_VERSION); gclient sync
+	mv build/v8 $(AUTOV8_DIR)
 
 $(AUTOV8_LIB): $(AUTOV8_DIR)
-	env CXXFLAGS=-fPIC $(MAKE) -C build/v8-$(AUTOV8_VERSION) dependencies
-	env CXXFLAGS=-fPIC $(MAKE) -C build/v8-$(AUTOV8_VERSION) native
+	env CXXFLAGS=-fPIC CFLAGS=-fPIC $(MAKE) -C build/v8-git-mirror-$(AUTOV8_VERSION) native
 	test -f $(AUTOV8_OUT)/libv8_base.a || \
 		ln -s $(abspath $(AUTOV8_OUT)/obj.target/tools/gyp/libv8_base.a) \
 			$(AUTOV8_OUT)/libv8_base.a
 	test -f $(AUTOV8_OUT)/libv8_snapshot.a || \
 		ln -s $(abspath $(AUTOV8_OUT)/obj.target/tools/gyp/libv8_snapshot.a) \
 			$(AUTOV8_OUT)/libv8_snapshot.a
+	test -f $(AUTOV8_OUT)/libv8_libbase.a || \
+		ln -s $(abspath $(AUTOV8_OUT)/obj.target/tools/gyp/libv8_libbase.a) \
+			$(AUTOV8_OUT)/libv8_libbase.a
+	test -f $(AUTOV8_OUT)/libv8_libplatform.a || \
+		ln -s $(abspath $(AUTOV8_OUT)/obj.target/tools/gyp/libv8_libplatform.a) \
+			$(AUTOV8_OUT)/libv8_libplatform.a
+	test -f $(AUTOV8_OUT)/libicui18n.a || \
+		ln -s $(abspath $(AUTOV8_OUT)/obj.target/third_party/icu/libicui18n.a) \
+			$(AUTOV8_OUT)/libicui18n.a
+	test -f $(AUTOV8_OUT)/libicuuc.a || \
+		ln -s $(abspath $(AUTOV8_OUT)/obj.target/third_party/icu/libicuuc.a) \
+			$(AUTOV8_OUT)/libicuuc.a
+	test -f $(AUTOV8_OUT)/libicudata.a || \
+		ln -s $(abspath $(AUTOV8_OUT)/obj.target/third_party/icu/libicudata.a) \
+			$(AUTOV8_OUT)/libicudata.a
 
 include Makefile
 
-CCFLAGS += -I$(AUTOV8_DIR)/include
+CCFLAGS += -I$(AUTOV8_DIR)/include -I$(AUTOV8_DIR)
 # We're gonna build static link.  Rip it out after include Makefile
 SHLIB_LINK := $(filter-out -lv8, $(SHLIB_LINK))
--- a/plv8.cc
+++ b/plv8.cc
@@ -6,6 +6,8 @@
  *-------------------------------------------------------------------------
  */
 #include "plv8.h"
+#include "libplatform/libplatform.h"
+
 #include <new>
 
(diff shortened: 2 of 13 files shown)


mgcrea:
This PR got [some pulbicity](http://bonesmoses.org/2015/04/10/pg-phriday-functions-and-addressing-json-data/?utm_source=postgresweekly&utm_medium=email#comment-37532):

> Unfortunately plv8 development seems currently stalled and not very responsive towards community contributions — the pull request on github containing modifications to support jsonb and an up-to-date V8 version including ES6 features is lying around uncommented and ignored now for over three weeks…

I do have big hopes for JSON-based workflows in PostgreSQL. So... Big :+1:!

JerrySievert:
+1

Moving to a modern v8 is a huge step, iojs has managed to move forward quite a bit. Allowing for es6 could also be huge.

JerrySievert:
This pull request now includes full JSONB support - can someone with merge access please review and consider merging (or at least comment on the issue?)

thanks!

JerrySievert:
@umitanuki can you take a look at this please?

thanks much!

umitanuki:
Sure, give me couple of days.

rafaelveloso:
Very much looking forward to having `JSONB` support. Any idea on when will this be merged?

JerrySievert:
@umitanuki any updates?  really excited to see this merged (and the jsonb support it brings).

thanks!

Makefile.v8
```diff
@@ -22,21 +26,35 @@ plv8_config.h: $(AUTOV8_LIB)
...
 	mkdir -p build
-	curl -L 'https://github.com/v8/v8/archive/$(AUTOV8_VERSION).tar.gz' \
-		| tar zxf - -C build
+	cd build; fetch v8; cd v8; git checkout $(AUTOV8_VERSION); gclient sync
+	mv build/v8 $(AUTOV8_DIR)
 
 $(AUTOV8_LIB): $(AUTOV8_DIR)
-	env CXXFLAGS=-fPIC $(MAKE) -C build/v8-$(AUTOV8_VERSION) dependencies
-	env CXXFLAGS=-fPIC $(MAKE) -C build/v8-$(AUTOV8_VERSION) native
+	env CXXFLAGS=-fPIC CFLAGS=-fPIC $(MAKE) -C build/v8-git-mirror-$(AUTOV8_VERSION) native -j12
```
umitanuki: This -j12 is not good.  The j option should be given from the command line and be propagated, as it depends on the compile environment how many parallel jobs are acceptable.

Makefile.v8
```diff
@@ -22,21 +26,35 @@ plv8_config.h: $(AUTOV8_LIB)
 
 $(AUTOV8_DIR):
 	mkdir -p build
-	curl -L 'https://github.com/v8/v8/archive/$(AUTOV8_VERSION).tar.gz' \
-		| tar zxf - -C build
+	cd build; fetch v8; cd v8; git checkout $(AUTOV8_VERSION); gclient sync
```
umitanuki: Can you tell me why we need this gclient now?

umitanuki:
We need regression tests to cover the new GUC as well as ES6 feature.

Makefile.v8
umitanuki: I'd love to eliminate this kind of dependency as much as possible to make it smooth for any packagers.

umitanuki:
Please update supported v8 version in README.md

plv8.cc
```diff
@@ -6,6 +6,8 @@
  *-------------------------------------------------------------------------
  */
 #include "plv8.h"
+#include <libplatform/libplatform.h>
```
umitanuki: If it's from v8 header, use double quotes

plv8.cc
```diff
@@ -428,10 +461,13 @@ CallFunction(PG_FUNCTION_ARGS, plv8_exec_env *xenv,
 			args[i] = ToValue(fcinfo->arg[i], fcinfo->argnull[i], &argtypes[i]);
 	}
 
+	Local<Object> recv = Local<Object>::New(plv8_isolate, xenv->recv);
 	Local<Function>		fn =
-		Local<Function>::Cast(xenv->recv->GetInternalField(0));
+		Local<Function>::Cast(recv->GetInternalField(0));
+		//Local<Function>::Cast(xenv->recv->GetInternalField(0));
```
umitanuki: Clean up these comments

JerrySievert:
@umitanuki thanks for looking!  i'll leave it to @schwebke for those changes - i'm most interested in the jsonb support i merged into this pull request.

umitanuki:
@JerrySievert I really should have responded to your pull request much earlier, but I would be going to add JSONB support first and work on this as it seems cleaner.  In any case, we need JSOB regression tests as well.

JerrySievert:
@umitanuki let me know what you want me to do - i can separate out my jsonb pull request again, or keep it here.  jsonb regression is simple, let me know if you want it here or in a separate request.

happy to oblige in any way!

JerrySievert:
@umitanuki btw, will you be at jsconf?

umitanuki:
@JerrySievert let's separate jsonb and add test cases.

JerrySievert:
Sounds good. I'll add test cases in the morning and resubmit my original pull request. 

Thanks!

Makefile.v8
schwebke: Using "make dependencies" in the v8 source tree instead of gclient leads to build errors (https://code.google.com/p/v8/issues/detail?id=3835#c11). Using gclient avoids these, and is currently the only way of building v8 which is supported by the v8 project.

schwebke:
@umitanuki
Thanks for the review. I've cleaned up the style issues you commented on. For gclient see the comment above.

With regards to regression tests I would need some help.

@JerrySievert 
It would be very nice if you could cherry pick the jsonb support from this branch, so it would not cause merge conflicts.

umitanuki:
@schwebke if this depot_tool is necessary to build static link, why don't you add the git clone for the tool in Makefile.v8 as well?

umitanuki:
@schwebke The JSONB support would have conflict with this change by nature, but that's easily fixable.

schwebke:
@umitanuki depot_tools are necessary for any kind of build of a recent v8 version (and already part of travis), and not specific to static linkage.

I've added them to static build as convenience.

umitanuki:
Got it.  As far as this local-fetch of gclient in Makefile.v8 doesn't break system that already has gclient in system path or somewhere, it looks good to me.  JSONB is merged in master, so I'd appreciate if you rebase this change on it.

umitanuki:
@schwebke how can I help you as to regression tests?  Have you ever run make installcheck?  I'm ok with a few additional test cases to cover the GUC change to be effective and some use of ES6 specific script.

schwebke:
@umitanuki I've added two basic tests (one for ES6 features enabled by default in the current V8, and one for potential breaking ES6 features which need the GUC). This PR is now rebased, and make installcheck passes all tests for me.

umitanuki:
Thanks!  I failed to build v8 on my mac as well as ubuntu 12.  Which platform did you build it?

schwebke:
@umitanuki on Ubuntu 14.04; 12.04 will be a hassle, 'cause v8 assumes many deps like git, python, gcc, ... to be not as outdated as the stock versions in 12.04.

umitanuki:
Merged.  Thanks!

BTW, I had wanted to support SRF with yield statement when it becomes official language spec.  Any idea to hack it?

JerrySievert:
failed to build on mac, which is not good:

```
Undefined symbols for architecture x86_64:
  "std::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> >::str() const", referenced from:
      std::string* v8::base::MakeCheckOpString<int, int>(int const&, int const&, char const*) in libv8_libbase.a(logging.o)
      std::string* v8::base::MakeCheckOpString<long, long>(long const&, long const&, char const*) in libv8_libbase.a(logging.o)
      std::string* v8::base::MakeCheckOpString<long long, long long>(long long const&, long long const&, char const*) in libv8_libbase.a(logging.o)
      std::string* v8::base::MakeCheckOpString<unsigned int, unsigned int>(unsigned int const&, unsigned int const&, char const*) in libv8_libbase.a(logging.o)
      std::string* v8::base::MakeCheckOpString<unsigned long, unsigned long>(unsigned long const&, unsigned long const&, char const*) in libv8_libbase.a(logging.o)
```

looks like the stdlib isn't being included in the link - this should get fixed asap.

can we reopen this issue until this is fixed?

JerrySievert:
FYI, this in Makefile.v8 fixes the compilation on OS X:

```
CCFLAGS += -I$(AUTOV8_DIR)/include -I$(AUTOV8_DIR) -stdlib=libstdc++
# We're gonna build static link.  Rip it out after include Makefile
SHLIB_LINK := $(filter-out -lv8, $(SHLIB_LINK)) -stdlib=libstdc++
```

not sure how that would affect ubuntu, i can test on 14.04, but am at JSConf and don't want to download an ubuntu 12.04 on the already overtaxed network.

schwebke:
@JerrySievert 

it might be even worse:
https://blog.inventic.eu/2014/01/qt-on-osx-maverick-undefined-symbols-for-architecture-x86_64/

It's possible the C++ lib is not just missing, but two different versions (-stdlib=libstdc++ and -stdlib=libc++) are mixed up. Could you check which C++ flags the V8 build is using?

JerrySievert:
Yeah, I read that too. Still tracking down but conference wifi makes things a little slower.

JerrySievert:
io.js manages to handle the dependencies fairly well - we can check that portion out, or i can see who i can ask on the project here.

JerrySievert:
all of the tests for mac are checking against libstdc++, so it appears to be something in the config upstream (plv8).

JerrySievert:
the issue occurs right at the end of the linking process with plv8 - so we will need to make a decision as to which stdlib to link with in the Makefile.v8.  unfortunately, i don't have easy access to the rest of the supported platforms to make determinations as to which to link against in all cases.

umitanuki:
I guess the issue is not unique to plv8.  I recently read somewhere the deep learning library Caffe installation also suggests building libstdc++ from brew by adding some env var to choose clang not gcc.

Wait, is it that v8 builds with its own libstdc++ and we are not using it?

JerrySievert:
@umitanuki from what i'm understanding, v8 builds on its own and expects any linking application (plv8 in this case), to correctly link to the system native libstdc++ library (which on OS X is libstdc++).

i could be wrong, looking for more confirmation.

JerrySievert:
also, it seems that even though OS X switched to libc++ as the default, it still ships with libstdlic++ (i was wrong above for recent versions of OS X).

PR #119, opened 2015-05-24
Add JSONB support

JerrySievert:
This adds `JSONB` support into `plv8_type.cc` for conversion to and from `JSONB`.  This is not any lighter weight with `plv8` than the existing `JSON`, but users of `JSONB` will get those performance gains elsewhere.


JerrySievert:
@umitanuki here you go!  let me know if you see any issues!

umitanuki:
Merged with some edits.  Thanks!

PR #123, opened 2015-06-04
fix osx compilation

JerrySievert:
this should fix OS X compilation for both `make static` and `make` with v8 already installed.

i tested locally, and this only affects OS X for me - please test for yourself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,17 @@ ifdef V8_SRCDIR
 override CPPFLAGS += -I$(V8_SRCDIR) -I$(V8_SRCDIR)/include
 endif
 
+ifeq ($(OS),Windows_NT)
+	# noop for now, it could come in handy later
+else
+	UNAME_S := $(shell uname -s)
+	ifeq ($(UNAME_S),Darwin)
+		CCFLAGS += -stdlib=libstdc++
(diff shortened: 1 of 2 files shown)


JerrySievert:
@umitanuki please let me know if this works for you, and if it doesn't you OS parameters and i will try to fix ASAP.

umitanuki:
Sure, will try soon.

Makefile
```diff
@@ -64,6 +64,17 @@ ifdef V8_SRCDIR
 override CPPFLAGS += -I$(V8_SRCDIR) -I$(V8_SRCDIR)/include
 endif
 
+ifeq ($(OS),Windows_NT)
```
umitanuki: Not sure why you want to silently skip it.  What would be here in case of mingw?  I know some people build plv8 in it.

JerrySievert: i hadn't heard of any compilation errors except on OS X (i don't have mingw, even in a vm).

i the check there just in case there are flags that need to be added for windows environments.  for now, since there weren't any, i left it as a no-op.

umitanuki: Ah ok, this is to patch flags only on osx.

JerrySievert: correct

PR #127, opened 2015-07-10
Crashfix

### LINKED ISSUE #126: CRASH - with new version of PLV8
```
CREATE OR REPLACE FUNCTION crash_test ( ) RETURNS BOOL
AS $$
  plv8.elog(WARNING, new ArrayBuffer(8));
  return true;
$$ LANGUAGE plv8 IMMUTABLE;
```

```sql
SELECT crash_test();
```

Error:

```
#
# Fatal error in ../src/runtime/runtime-typedarray.cc, line 65
# CHECK(V8::ArrayBufferAllocator() != NULL) failed
#

==== C stack trace ===============================

 1: ??
 2: ??
 3: ??
 4: ??
LOG:  server process (PID 30469) was terminated by signal 4: Illegal instruction
DETAIL:  Failed process was running: select crash_test();
```

JerrySievert:
there is a crash due to a failed assertion when trying to use an ArrayBuffer (or anything that uses an ArrayBuffer internally).

the issue is due to `plv8` not providing an allocation handler.

this pull request adds an allocation handler which uses `palloc` and `pfree`, and sets it up before initialization.

there is not an additional regression test for the crash.

closes #126 

/cc @umitanuki

### DIFF
--- a/expected/arraybuffer.out
+++ b/expected/arraybuffer.out
@@ -6,7 +6,7 @@ AS $$
 $$;
 SELECT crash_test();
 WARNING:  Allocating: 8
- crash_test
+ crash_test 
 ------------
  t
 (1 row)
--- a/expected/arraybuffer.out
+++ b/expected/arraybuffer.out
@@ -1,0 +1,12 @@
+CREATE FUNCTION crash_test() RETURNS boolean
+LANGUAGE plv8 IMMUTABLE STRICT
+AS $$
+  var a = new ArrayBuffer(8);
+  return true;
+$$;
+SELECT crash_test();
+WARNING:  Allocating: 8
+ crash_test
+------------
+ t
+(1 row)
--- a/expected/arraybuffer.out
+++ b/expected/arraybuffer.out
@@ -10,3 +10,4 @@ WARNING:  Allocating: 8
 ------------
  t
 (1 row)
+
--- a/plv8.cc
+++ b/plv8.cc
@@ -134,6 +134,16 @@ static plv8_exec_env		   *exec_env_head = NULL;
 extern const unsigned char coffee_script_binary_data[];
 extern const unsigned char livescript_binary_data[];
 
+class Plv8ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
+ public:
+  virtual void* Allocate(size_t length) {
+    void* data = AllocateUninitialized(length);
+    return data == NULL ? data : memset(data, 0, length);
+  }
+  virtual void* AllocateUninitialized(size_t length) { return palloc(length); }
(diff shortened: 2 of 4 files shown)


plv8.cc
```diff
@@ -134,6 +134,16 @@ static plv8_exec_env		   *exec_env_head = NULL;
 extern const unsigned char coffee_script_binary_data[];
 extern const unsigned char livescript_binary_data[];
 
+class Plv8ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
+ public:
+  virtual void* Allocate(size_t length) {
+    void* data = AllocateUninitialized(length);
+    return data == NULL ? data : memset(data, 0, length);
+  }
+  virtual void* AllocateUninitialized(size_t length) { return palloc(length); }
```
haradh1: Shouldn't it be wrapped by PG_TRY block to suppress sigjmp in case of out of memory?

umitanuki: Shouldn't it be wrapped by PG_TRY block to suppress sigjmp in case of out of memory?

expected/arraybuffer.out
```diff
@@ -0,0 +1,13 @@
+CREATE FUNCTION crash_test() RETURNS boolean
+LANGUAGE plv8 IMMUTABLE STRICT
+AS $$
+  var a = new ArrayBuffer(8);
+  return true;
+$$;
+SELECT crash_test();
+WARNING:  Allocating: 8
```
umitanuki: Why is this WARNING emit?  Is it v8 internal stuff??

JerrySievert: no, it's an artifact of me not getting the correct expected file from my vm - fixing now.

plv8.cc
JerrySievert: i can do that if you think it makes sense.  nowhere else in the code is using `PG_TRY` around a `palloc` though.

let me know, and i'm happy to add it.

umitanuki: No it's actually mandatory to prevent PG exception from leaking to C++ layer, I guess.  If you take a close look at other places including caller stack, we catch the PG exception and translate it to C++ pg_error exception before going back to C++, one way or another.  So I'd like you to do it.

JerrySievert:
i've added `PG_TRY()` as requested, and regression tests pass for me.

please let me know if there is anything else you would like changed.

umitanuki:
Should this be backported to 1.4?

umitanuki:
Merged to master (dab5fdf)

JerrySievert:
`ArrayBuffer` didn't exist in the version of v8 supported by 1.4

umitanuki:
ok

PR #144, opened 2016-02-13
update build

JerrySievert:
this should fix some travis issues, as well as OS X builds.


JerrySievert:
this includes some fixes for travis, as well as fixes for OS X.

PR #147, opened 2016-02-18
PG version number from server include dir

car3ca:
This should fix Makefile problems for versions under 9.4


JerrySievert:
awesome! merging this in - this leaves us a couple of travis failures, but we're getting really close.

thanks!

car3ca:
No problem!
I can't spend much more time on this issue this week, but I'll check it next week.
Thanks!

JerrySievert:
sounds good - progress is progress :)

PR #150, opened 2016-02-21
update travis tests

JerrySievert:
this pull request updates the following in travis:

* removes older versions of v8 that are no longer shipping with the v8 source tree (4.1 and lower)
* asks to run on the new *trusty* platform
* compiles with both `clang++` and `g++`

this should fix tests enough to start to be able to trust travis builds and include a badge.


JerrySievert:
all checks passing, merging.

PR #153, opened 2016-02-23
update changes, readme, and fix memory leak

### LINKED ISSUE #64: cursor.fetch leaks memory
```
https://t.co/ijdGoRWQji

Originally reported by Makoto YUI

It seems the object returned by Converter::ToValue is not garbage-collected in 
the loop.
```

Original issue reported on code.google.com by `umi.tan...@gmail.com` on 11 Apr 2013 at 7:29

JerrySievert:
this should fix a memory leak, and close #64.

### DIFF
--- a/Changes
+++ b/Changes
@@ -7,6 +7,7 @@ x.x.x       2015-xx-xx
             - GUC plv8.v8_flags for V8 engine initialization flags
             - Fixed crash in ArrayBuffer
             - Fixed encoding issues with elog
+            - Fixed memory leak in cursor fetch
 
 1.4.4       2015-05-26
             - Add jsonb type coercion in function boundary.
--- a/Changes
+++ b/Changes
@@ -1,12 +1,12 @@
 Revision history for plv8
 
 x.x.x       2015-xx-xx
-            - Adaptions for current V8 version, e.g. 4.1.0.x
+            - Adaptions for current V8 version, e.g. 4.1.0.x - 4.3.66
               (see http://strongloop.com/strongblog/node-js-v0-12-c-apis-breaking/
               for an overview of the V8 changes which break existing code bases)
             - GUC plv8.v8_flags for V8 engine initialization flags
-            - Extended error object when an error is thrown from Postgres
             - Fixed crash in ArrayBuffer
+            - Fixed encoding issues with elog
 
 1.4.4       2015-05-26
             - Add jsonb type coercion in function boundary.
--- a/README.md
+++ b/README.md
@@ -10,9 +10,10 @@ REQUIREMENT
 
 plv8 is tested with:
 
-- PG: version 9.3 and 9.4 (maybe older/newer are allowed)
-- V8: version 4.3.66
+- PG: version 9.2, 9.3, 9.4 and 9.5 (maybe older/newer are allowed)
+- V8: version 4.2 - 4.3.66 (as of v1.5.0)
 - g++: version 4.8.2
+- clang++
 
 It is also known to work with some older versions of gcc and v8.
 
--- a/plv8_func.cc
+++ b/plv8_func.cc
@@ -860,6 +860,8 @@ plv8_CursorFetch(const FunctionCallbackInfo<v8::Value> &args)
 			return;
 		}
 	}
+
+	SPI_freetuptable(SPI_tuptable);
 	args.GetReturnValue().Set(Undefined(plv8_isolate));
 }
 
--- a/plv8_func.cc
+++ b/plv8_func.cc
@@ -842,6 +842,8 @@ plv8_CursorFetch(const FunctionCallbackInfo<v8::Value> &args)
 		{
 			Handle<v8::Object>	result = conv.ToValue(SPI_tuptable->vals[0]);
 			args.GetReturnValue().Set(result);
+			SPI_freetuptable(SPI_tuptable);
+
 			return;
 		}
 		else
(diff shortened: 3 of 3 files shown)


JerrySievert:
@eugwne any chance you can take a look at this?  it has dropped my memory usage to almost not be noticeable.

eugwne:
If I'm not mistaken SPI_freetuptable should be done for each SPI_cursor_fetch and might be better to use something like scoped object with SPI_freetuptable inside destructor. If variable can be accessed outside transaction than CreateTupleDescCopy FreeTupleDesc might be used inside Converter to store m_tupdesc. Tried to implement this in https://github.com/eugwne/plv8/commit/27b27ec448a0b6005b723207b5aa33827a50a879

eugwne:
checked this code
```javascript
do $$

for (var k=0; k<10; ++k){
var plan = plv8.prepare('select v from generate_series(1,50) as v');
var cursor = plan.cursor();
while (cursor.fetch()){}
cursor.close();
plan.free();
}
$$ language plv8
```
and result:
```
TopMemoryContext: 102704 total in 14 blocks; 33016 free (113 chunks); 69688 used
  TopTransactionContext: 8192 total in 1 blocks; 6224 free (2 chunks); 1968 used
    SPI Exec: 8192 total in 1 blocks; 8160 free (0 chunks); 32 used
    SPI Proc: 8192 total in 1 blocks; 7360 free (1 chunks); 832 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
      SPI TupTable: 8192 total in 1 blocks; 6768 free (0 chunks); 1424 used
  CFuncHash: 8192 total in 1 blocks; 1680 free (0 chunks); 6512 used
  PLv8 Procedures: 25392 total in 2 blocks; 3744 free (0 chunks); 21648 used
  TableSpace cache: 8192 total in 1 blocks; 3216 free (0 chunks); 4976 used
...
```

just one more SPI_freetuptable(SPI_tuptable)  needs here https://github.com/JerrySievert/plv8/blob/master/plv8_func.cc#L863

JerrySievert:
good catch thanks!

eugwne:
Tried to check this path https://github.com/JerrySievert/plv8/blob/master/plv8_func.cc#L837 but got exceptions and crushes for this:
```
--exceptions
create or replace function pg_temp.ex() returns integer as $$
begin
raise exception 'custom error';
end
$$ language plpgsql

--crush 
create or replace function pg_temp.ex() returns integer as $$
throw 'custom error';
$$ language plv8

do $$
for (var k=0; k<10; ++k){
var plan = plv8.prepare('select pg_temp.ex()');
var cursor = plan.cursor();

try{
	while (cursor.fetch()){}
}catch(e){}
cursor.close();
plan.free();
}
$$ language plv8
```

JerrySievert:
ok, i will try to track that down

JerrySievert:
well, the good news is that this caused a crash in creating a plan before this pull request.

```
SPI_cursor_open_internal(name=<unavailable>, plan=0x0000000000000000, paramLI=<unavailable>, read_only='\0')
```

plan (NULL)  is being dereferenced.

JerrySievert:
i found (and can fix) the crash, but am undecided what error message to return back:

```
	/* XXX: Add plan validation */
```

JerrySievert:
```
	if (plan == NULL) {
		StringInfoData	buf;

		initStringInfo(&buf);
		appendStringInfo(&buf,
				"plan unexpectedly null");
		throw js_error(pstrdup(buf.data));
	}
```

eugwne:
It would be great to get  exception that was thrown from pg_temp.ex

eugwne:
It seems to me that it looks like this:
1) when cursor.fetch() inside throws an error it goes through pg error, and if are in subtransaction block   we must cancel it, if not than cancel transaction and stop execution
2) exception goes js->pg->js, we do not stop execution, catch js error and continue
3) pg error thrown, we continue and try execute plv8.prepare and eventually get plan NULL (queries SPI will not work in current transaction)

I think this code should stop on  _while (cursor.fetch()){}_ line
```
create or replace function pg_temp.ex() returns integer as $$
throw 'custom error';
$$ language plv8

do $$
var plan = plv8.prepare('select pg_temp.ex()');
var cursor = plan.cursor();
try{
    while (cursor.fetch()){}
}catch(e){
    plv8.elog(NOTICE, e)
}
cursor.close();
plan.free();
plv8.elog(NOTICE, 'code still executes')
$$ language plv8

NOTICE:  Error: custom error
NOTICE:  code still executes
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

It looks like it needs another issue/pull request

eugwne:
"plan unexpectedly null" is a good message for such an error

JerrySievert:
agreed on the exception needing to bubble up correctly, but going to reserve that for another ticket.

going to merge this, and move on for now.

PR #174, opened 2016-04-16
fix resource owner and context

JerrySievert:
this fixes resource owner issues and context issues.

much better fix moving forward.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -142,6 +142,7 @@ class Plv8ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
   }
   virtual void* AllocateUninitialized(size_t length) {
 			void *data = NULL;
+			MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
 
 			PG_TRY();
 			{
@@ -156,6 +157,8 @@ class Plv8ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
 			return data;
 	}
   virtual void Free(void* data, size_t) {
+			MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
+
 			PG_TRY();
 			{
 					pfree(data);
--- a/plv8.cc
+++ b/plv8.cc
@@ -154,6 +154,8 @@ class Plv8ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
 			}
 			PG_END_TRY();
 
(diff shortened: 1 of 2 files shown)


JerrySievert:
@ruimarinho this should fix any issues moving forward.  about to close my day, will try to work on other issues tomorrow.

ruimarinho:
Interesting - my experience with resource owner crashes are not using plv8.execute/plv8.subtransaction. I'll test this patch.

JerrySievert:
i think this is a a nascent issue, and may not be as exposed due to garbage collection.

there's no reason why it shouldn't be changed regardless

ruimarinho:
I have just confirmed the resource owner issue still exists on v1.5.2.

ruimarinho:
Also confirmed as not fixed on v1.4.7.

JerrySievert:
do you have a simple test case for this that we can add to regressions?

we have one simple regression already: https://github.com/plv8/plv8/blob/master/sql/currentresource.sql#L1-L9

ruimarinho:
Not yet, but I can reproduce it and it's inside this query (where _ and assign come from the previously discussed memory context example issue):

```sql
UPDATE foos
SET bar = _.assign(bar, '{"prop":1}')
WHERE "id" = '7a19fee6-ed96-4ddd-bcc8-6916a30e2005'
```

JerrySievert:
hm. the odd part of that is that _ isn't using a `plv8.execute`.

ruimarinho:
I'm on this now.. do you have spare time for some debugging? I'm on irc (same handle as github's).

JerrySievert:
that is awesome - unfortunately my tuesdays are always booked.  i can help tomorrow, though.

ruimarinho:
No worries :) I'll try to join tomorrow then.

JerrySievert:
as a heads up, i'm "jerrysv" on freenode (sv is the international symbol for "sievert"), and i'm online about 14 hours/day - feel free to message me, i am not always available (i'm in a ton of meetings), but can at least asynchronously reply!

JerrySievert:
@christopherberg released 1.4.8, should be the last fix for this, i hope.

ChristophBerg:
Debian packages built, thanks for the notice!

JerrySievert:
thanks @ChristophBerg ! you rock!

ruimarinho:
Confirmed as fixed on 1.4.8 and 1.5.3. Awesome work @JerrySievert!

Issue #184, opened 2016-06-08
Compile Failure


dl748:
Running make,

g++  6.1.1
libv8 5.2.258

Error for every .cc file,

error: ‘namespace_’ does not name a type
 #define namespace namespace_

error: ‘using_’ does not name a type
 #define using  using_

error: ‘typename_’ does not name a type
 #define typename  typename_

So I commented them out in every .cc file and the compiling worked for every file.

When linking after compile, failed with

/usr/bin/ld: cannot find -lv8_libplatform
collect2: error: ld returned 1 exit status

Commented out

SHLIB_LINK += -lv8_libplatform
to
## SHLIB_LINK += -lv8_libplatform

Finally compilation was successful.

Also, no "make test" ?

JerrySievert:
thanks for the update.  can you verify that those were part of v8?  i'm not finding them in the plv8 project itself.

if that's the case, it's an upstream issue that we can only track, and not fix directly, which is unfortunate.

dl748:
They are not apart of v8, but apart of plv8, plv8.cc lines 14-18 are an example.  I think its a problem when compiling with G++ instead of CLang.

JerrySievert:
on the `make test`, postgres extensions inherit the postgres testing strategy via `pg_config`, so as noted in the README, the method for testing is:

```
$ make installcheck
```

this runs the regression tests that are in `sql`, and validates them against `expected`.

i am trying to replicate your issue right now, and on fix will add `gcc-6.1` into the test matrix.  previously, we had not been compiling against this target in our ci.

JerrySievert:
i'm not even able to compile that version of v8 - can you tell me what OS you are compiling on, so i can emulate your environment?  i'm running on ubuntu 16.04 against `g++-6.1`.

dl748:
Fedora, I have a package for v8, but I can also compile it from source.

gcc --version
g++ (GCC) 6.1.1 20160510 (Red Hat 6.1.1-2)
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

v8.x86_64                                                   1:5.2.258

ls -l /usr/lib/libv8*

/usr/lib64/libv8_libplatform.a
/usr/lib64/libv8.so -> libv8.so.5
/usr/lib64/libv8.so.5

dl748:
[...]

PR #188, opened 2016-10-07
Remove #defines papering over C++ keywords

ChristophBerg:
These cause build failures with gcc-6.

This fixes the 1.4 branch for #184.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -9,12 +9,6 @@
 #include <new>
 
 extern "C" {
-#define delete		delete_
-#define namespace	namespace_
-#define	typeid		typeid_
-#define	typename	typename_
-#define	using		using_
-
 #if PG_VERSION_NUM >= 90300
 #include "access/htup_details.h"
 #endif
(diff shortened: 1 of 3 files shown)


JerrySievert:
you sir, are amazing.  waiting for tests to pass, then will merge - and try to move it forward to 1.5 and 2.0.

JerrySievert:
hm. all checks failing - not sure why, will try to look into it tonight or tomorrow, depending on how much free time i get.

issue seems to be related to checkout of v8 itself.

JerrySievert:
it looks like `3.18` is no longer available in the git repo.  might look at installing the OS's libv8 for now, for 1.4, can't wait to see debian based os's moving to a more modern v8, but you know this already :)

ChristophBerg:
Fwiw, this patch has passed compiling+regression testing on apt.postgresql.org for Debian sid+jessie+wheezy and Ubuntu xenial+trusty (precise's gcc doesn't have c++11), so removing the #defines at least didn't break anything on Linux. (And re the newer v8, yes that would be nice, but that's up to other people to fix...)

JerrySievert:
since i know you tested this, i'm going to merge and try to get some time to fix the CI on this when i'm not traveling.

PR #235, opened 2017-09-03
Update plv8.md with formatting and another example

deinspanjer:
* Add subheadings under the Utility section to make it easier to find information about specific functions (I skimmed over find_function twice before finally remembering the name and doing a string search for it)
* Quick example of plv8.elog

### DIFF
--- a/doc/plv8.md
+++ b/doc/plv8.md
@@ -461,10 +461,9 @@ PL/v8 provides the following utility built-in functions.
 - `WARNING`
 - `ERROR`
 
-```javascript
-var msg = 'world';
-plv8.elog(DEBUG1, 'Hello',`${msg}!`);
-```
+    var msg = 'world';
+    plv8.elog(DEBUG1, 'Hello',`${msg}!`);
+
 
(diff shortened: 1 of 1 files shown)


doc/plv8.md
```diff
@@ -459,13 +461,20 @@ PL/v8 provides the following utility built-in functions.
 - `WARNING`
 - `ERROR`
 
+```javascript
+var msg = 'world';
```
JerrySievert: unfortunately, this won't work - we're published on pgxn, which does not support gfm.  please check how other examples are formatted.

deinspanjer: Ahh, sorry I was only thinking about github syntax.  I'll clean that up right now.

JerrySievert: don't worry, you're not alone :) it was @theory who put in the PR to fix the previous formatting to work correctly with pgxn.  thanks much for the update, I'll finish the review in just a little bit!

JerrySievert:
awesome, thanks!

PR #247, opened 2017-12-15
Build system and Fedora build

praiskup:
FYI, I needed to change those to correctly build for Fedora 27 (we use
CXXFLAGS for setting proper debuginfo flag).  Any comment is welcome.


Makefile
```diff
@@ -62,7 +60,7 @@ endif
 
 OPTFLAGS = -O2 -std=c++11 -fno-rtti
 
-CCFLAGS = -Wall $(OPTFLAGS) $(OPT_ENABLE_DEBUGGER_SUPPORT)
+CCFLAGS = -Wall $(CXXFLAGS) $(OPTFLAGS) $(OPT_ENABLE_DEBUGGER_SUPPORT)
```
JerrySievert: unfortunately, this will break builds for most other platforms.

praiskup: Thanks for having a look.  You probably meant the -lv8_libplatform removal, which is indeed needed (also on Fedora, I was just mistaken by Fedora bug https://bugzilla.redhat.com/show_bug.cgi?id=1517657).

JerrySievert: yeah, I meant to comment on the `v8_libplatform` line.

praiskup:
I pushed one commit more adding 'make test' functionality, that tests without the need to do 'make install' first (this is useful in CI environments where you don't have root access).  For x86_64/i686 it works fine, but problems are on armv7hl architecture: https://kojipkgs.fedoraproject.org//work/tasks/9093/23779093/build.log

Please let me know whether I should iterate on the 'make test' feature, or you will, or I should keep that downstream-only.

praiskup:
Ping.

JerrySievert:
keep downstream only for now, I would rather stick with standard postgres testing methodologies

praiskup:
ACK, plv8 is probably EOL in Fedora anyways.

praiskup:
> I would rather stick with standard postgres testing methodologies

Add "installcheck" -- Note that to test the built plugin -- which is to be installed into `/usr` -- you need to `make install` first.  That implies you have to be root, or you can not test at all.  Sane downstream packages never give root access to build scripts....   Just to let you know what was my motivation.

PR #278, opened 2018-05-10
get ready for 2.3.4

JerrySievert:
* upgrade creation scripts
* fix pre-build for windows
* add execution timeout for windows
* fix PGXN path


JerrySievert:
@kjunichi would you be willing to try the windows build of this?  I removed most of the prebuild parts, and they should auto-build now as part of bootstrap.

PR #280, opened 2018-06-01
Makefile.shared: don't install versioned module

praiskup:


JerrySievert:
is there some reason why you want to get rid of all versioning for modules for plv8?

praiskup:
Uh, sorry, it's just non-standard thing;  and it broke our specfile (that can be fixed on our side).  Do we need the version string in so file?

JerrySievert:
```
/usr/local/lib/postgresql/postgis-2.4.so
/usr/local/lib/postgresql/postgis-2.5.so
/usr/local/lib/postgresql/rtpostgis-2.5.so
/usr/local/lib/postgresql/postgis_topology-2.4.so
/usr/local/lib/postgresql/postgis_topology-2.5.so
/usr/local/lib/postgresql/rtpostgis-2.4.so
```

seems to have become a de facto standard for complicated modules that rely on other code that has braking changes between versions.

JerrySievert:
fyi, change also breaks plv8 completely - see the travis checks.

praiskup:
Nope, postgis is the only exception I'm aware of (all those you cited are from postgis);  and as such it complicated Fedora distro upgrades (and following `pg_upgrade`) a lot.   Now I finally worked-around that (by building the so file 3x, twice against new server, once against old server).

praiskup:
Just try to google for the "pg_upgrade" and "postgis" headaches.  Definitely not a standard :-).

> fyi, change also breaks plv8 completely - see the travis checks.

To the topic -- I'd try to fix the travis if the change is otherwise acceptable...

PR #311, opened 2018-11-28
bump v8 to 7.0.276.40

pkit:
7.0.276.40 is the latest chromium stable


JerrySievert:
I already have a branch that moves to 7.0.276.20, as well as activating `BigInt` support, there is an open comment thread on the mailing list, so going to close this.

I do appreciate the PR, but this is missing a few things that are already addressed:

* query timeouts
* windows support

JerrySievert:
fyi, the current work in progress is here: https://github.com/jerrysievert/plv8/tree/r3.0alpha

PR #319, opened 2019-02-15
Include 2.1.2 in generate_upgrade.sh for sql upgrade scripts

JohnHVancouver:
I assume 2.1.2 was missed accidentally so this PR is to add it to generate_upgrade.sh to support ALTER EXTENSION UPDATE from 2.1.2 to newer versions.

Note: 2.1.2 is missing in earlier tags as well.


JerrySievert:
hm. I'm pretty sure that `2.1.2` simply was not released, so wondering if it's worth it to have this.  my memory says it was skipped over, but I could be wrong.

JohnHVancouver:
Interesting, what do you mean by not released?
The tag exists here:
https://github.com/plv8/plv8/tree/v2.1.2

JerrySievert:
as in no official released version.  I'm ok accepting the PR, but could you please add it to the other part that generates these?

https://github.com/plv8/plv8/blob/r3.0alpha/windows/generate_upgrade.ps1

thanks!

JohnHVancouver:
Sweet thanks!
I've included a new commit. Let me know if you would prefer me to squash it.

JerrySievert:
nah, I can squash on my end. thanks!

JerrySievert:
ah, it was `2.2.*` that was skipped over, knew it was somewhere around there.

PR #322, opened 2019-02-24
add pg 10 & 11 to travis - turns out to work

asah:
note that travis only provides major versions for pg 10 and 11

successful test runs:
https://travis-ci.com/asah/plv8


JerrySievert:
awesome, thanks!

PR #342, opened 2019-05-20
upgrade to v8 7.4.288.28

pkit:
partially fixes #340

PR #343, opened 2019-05-20
V8 7.4

### LINKED ISSUE #340: array spread will crash the server process
```
do $$ 
     return {
       next:() => this
     }
  };
  [...({})]; 
$$ language plv8
```
fixed here: https://github.com/v8/v8/commit/1c48d52bb1ee9bb28e146c60eda08cd4afaa5745

JerrySievert:
fixes #340 and supersedes #342.


JerrySievert:
@pkit I'll add the other changes, but this at least gets travis working.

pkit:
@JerrySievert note that id doesn;t really fix #340 just makes it easier to fix it later (OOM hook, isolate memory restrictions and a custom ArrayBuffer allocator are still needed)

JerrySievert:
hence the reason why I didn't merge it yet.  just wanted to get the "hard" part into a PR.

pkit:
btw, got a better code for the platform part:
```
diff --git a/plv8.cc b/plv8.cc
index 2e2eb5f..bec44f9 100644
--- a/plv8.cc
+++ b/plv8.cc
@@ -183,6 +183,8 @@ static int plv8_debugger_port;
 static int plv8_execution_timeout = 300;
 #endif
 
+static std::unique_ptr<v8::Platform> v8_platform = NULL;
+
 /*
  * We use vector instead of hash since the size of this array
  * is expected to be short in most cases.
@@ -300,8 +302,10 @@ _PG_init(void)
 #if (V8_MAJOR_VERSION == 4 && V8_MINOR_VERSION >= 6) || V8_MAJOR_VERSION >= 5
        V8::InitializeExternalStartupData("plv8");
 #endif
-       Platform* platform = platform::NewDefaultPlatform().release();
-       V8::InitializePlatform(platform);
+       if (!v8_platform) {
+               v8_platform = platform::NewDefaultPlatform();
+       }
+       V8::InitializePlatform(v8_platform.get());
        V8::Initialize();
        if (plv8_v8_flags != NULL) {
                V8::SetFlagsFromString(plv8_v8_flags, strlen(plv8_v8_flags));

```
it prevents re-init of Platform class on fork
probably that' why they introduced that change

pkit:
crap, ArrayBuffer allocator is tricky
if arrays are allocated in small increments (<=64 bytes) v8 will hard crash on nullptr from the allocator
i.e. probably need to call exit(1) on the backend pid in such cases

PR #345, opened 2019-05-23
Oom handling

pkit:
there were the following corner cases found:
- allocate two big arrays (but smaller than heap limit) in subsequent isolate runs: should not fail
- allocate array over the heap limit in small increments (less than 64 bytes): should throw error
- run array spread bug repro: should fail with error
- run array spread repro twice: should fail with error twice

right now all these cases are handled correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ AUTOV8_STATIC_LIBS = -lv8_base -lv8_snapshot -lv8_libplatform -lv8_libbase -lv8_
 export PATH := $(abspath $(AUTOV8_DEPOT_TOOLS)):$(PATH)
 
 SHLIB_LINK += -L$(AUTOV8_OUT) -L$(AUTOV8_OUT)/third_party/icu $(AUTOV8_STATIC_LIBS)
-V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_snapshot=true v8_use_external_startup_data=false is_clang=false use_custom_libcxx=false
+V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_snapshot=true v8_use_external_startup_data=false use_custom_libcxx=false
 
 ifndef USE_ICU
 	V8_OPTIONS += v8_enable_i18n_support=false
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@
 # structure in v8 which may be different from version to another, but user
 # can specify the v8 version by AUTOV8_VERSION, too.
 #-----------------------------------------------------------------------------#
-AUTOV8_VERSION = 7.0.302
+AUTOV8_VERSION = 7.4.288.28
 AUTOV8_DIR = build/v8
 AUTOV8_OUT = build/v8/out.gn/x64.release/obj
 AUTOV8_DEPOT_TOOLS = build/depot_tools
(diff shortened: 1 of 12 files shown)


plv8.cc
```diff
@@ -285,6 +302,18 @@ _PG_init(void)
 							NULL);
 #endif
 
+	DefineCustomIntVariable("plv8.memory_limit",
+							gettext_noop("Per-isolate memory limit in MBytes"),
+							gettext_noop("The default value is 256 MB"),
+							(int*)&plv8_memory_limit,
```
JerrySievert, line 324: 256MB might be a little low, especially when dealing with ETL.  I'd feel more comfortable att around 1GB.  while it's not as big of a deal with the direct JSONB conversions, there are still some inefficiencies that are possible.

Makefile
```diff
@@ -16,7 +16,7 @@ AUTOV8_STATIC_LIBS = -lv8_base -lv8_snapshot -lv8_libplatform -lv8_libbase -lv8_
 export PATH := $(abspath $(AUTOV8_DEPOT_TOOLS)):$(PATH)
 
 SHLIB_LINK += -L$(AUTOV8_OUT) -L$(AUTOV8_OUT)/third_party/icu $(AUTOV8_STATIC_LIBS)
-V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_snapshot=true v8_use_external_startup_data=false
+V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_snapshot=true v8_use_external_startup_data=false is_clang=false use_custom_libcxx=false
```
JerrySievert: not sure why the `is_clang=false`, macOS uses clang by default, not gcc, so this might be something better left for ninja to figure out.

expected/array_spread.out
```diff
@@ -0,0 +1,13 @@
+do $$
+  Object.prototype[Symbol.iterator] = function() {
+     return {
+       next:() => this
+     }
+  };
+  [...({})];
+$$ language plv8;
+FATAL:  plv8: OOM error in GC
+server closed the connection unexpectedly
```
JerrySievert: I wonder if this might be better served by throwing an error, since it is specifically a javascript OOM error, not a full server OOM.

```
throw js_error("OOM error");
```

pkit: this is what happens with `js_error` or any other non-fatal thing:
```
terminate called after throwing an instance of 'js_error'
2019-05-24 11:08:43.852 EEST [21345] LOG:  server process (PID 21389) was terminated by signal 6: Aborted
2019-05-24 11:08:43.852 EEST [21345] DETAIL:  Failed process was running: do $$ 
             return {
               next:() => this
             }
          };
          [...({})]; 
        $$ language plv8
        ;
2019-05-24 11:08:43.853 EEST [21345] LOG:  terminating any other active server processes
```

plv8.cc
pkit: it's only a default can be set to anything in the config file
testing oom with 1GB will take much longer

Makefile
pkit: ok, removed it

expected/array_spread.out
pkit: okay, found out another hook, that tremendously improved the robustness 
now it can safely throw `js_error`

PR #362, opened 2019-08-29
Update BUILDING.md

claytongulick:


JerrySievert:
thanks, merging - I'll update plv8.github.io as well.

note that the ci failure is a timeout, not an actual failure so I'm ignoring.

PR #368, opened 2019-10-09
fix doc: keep the $$ (it was replaced to $ currently)

zxdvd:
I found that all `$$` became `$` currently due to the string.replace of js treats `$$` specially (insert a single $) in some case.

MDN has detail explanation at [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter).

To disable this special behavior, just using a function to replace the new string.

### DIFF
--- a/contrib/web/index.js
+++ b/contrib/web/index.js
@@ -47,7 +47,7 @@ files.forEach((elem) => {
 });
 
 // replace the content of the template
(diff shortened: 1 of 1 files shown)


JerrySievert:
merged, thank you!

PR #369, opened 2019-10-29
uname -m can also be aarch64 for arm64 platforms

xsacha:


JerrySievert:
awesome, thanks for the PR!  I only have 6 arm platforms I can test against, so it's a bit of a weakness for me.

much appreciated!

Issue #373, opened 2019-11-22
PostgreSQL 12.1 and PLV8 2.3.13 => PostgreSQL crashes


JamesInform:
HI!

I am not sure if my issue is related to PostgreSQL or to the PLV8 extension. 
So I just post in case.

I encountered the following severe issue. I am running PostgreSQL 12.1 on Ubuntu. Postgresql is build from source by myself, but the same issue appears when I am using Pg12.1 with Postgresapp on Mac. (www.postgresapp.com)

I am using PLV8 Version 2.3.13.

Steps to reproduce:

1. Create a new database:
create database mytest;

2. Create PLV8 extension
create extension plv8;

3. Create a simple PLV8 function
CREATE OR REPLACE FUNCTION plv8_require()
 RETURNS void
 LANGUAGE plv8
AS 
$function$
	var test = 0;
$function$
;

4. Log out and then in again with a new session to the new database mytest;

5. Run the following set command
SET plv8.start_proc = plv8_require;

6. Run this simple plv8 script
do
$$
plv8.elog(NOTICE,'test');
$$ language plv8
;

7. Now the server goes down with the following log information:

Received signal 4 <unknown> 000107b765be

==== C stack trace ===============================

 [0x00010838ed74]
 [0x7fff5a988f5a]
 [0x000107ab4888]
 [0x000107b7383c]
 [0x000104b07fdc]
 [0x000104b09091]
 [0x000104699360]
 [0x0001049491ac]
 [0x00010733dfdf]
 [0x000104948d1d]
 [0x00010494871d]
 [0x000104947a51]
 [0x000104947121]
 [0x000104944810]
 [0x000104941df7]
 [0x00010488f3ff]
 [0x00010488e80d]
 [0x00010488d76a]
 [0x00010488b37f]
 [0x00010479b5d9]
 [0x7fff5a67a015]
[end of stack trace]
2019-11-22 17:29:05 CET [50424]: [13-1] user=,db=,app=,client= LOG:  server process (PID 15664) was terminated by signal 4: Illegal instruction: 4
2019-11-22 17:29:05 CET [50424]: [14-1] user=,db=,app=,client= DETAIL:  Failed process was running: do
	$$
	plv8.elog(NOTICE,'test');
	$$ language plv8
	
2019-11-22 17:29:05 CET [50424]: [15-1] user=,db=,app=,client= LOG:  terminating any other active server processes
[...]

Issue #370, opened 2019-11-04
segfault during plv8.find_function on CentOS - PG12


nonara:
**Issue**
Segfault crash during plv8.find_function - only happens for valid functions.

**Setup**
* CentOS Linux release 7.7.1908 (Core)
* PostgreSQL 12.0 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39), 64-bit

**Detail**
* Crash occurs on both 2.3.13 and 3.0-alpha (with both clang and g++ of varying versions up to the latest)
* Seems isolated to PostgreSQL 12. (PG-11 works)

**Log**
Nov 04 18:21:29 kernel: postmaster[11344]: segfault at 0 ip 00007f120e53d252 sp 00007fff298a84d0 error 6 in plv8-3.0alpha.so[7f120e492000+c16000]
Nov 04 18:21:29 kernel: Code: 24 80 00 00 00 48 89 c1 48 83 c0 08 48 c7 40 f8 00 00 00 00 48 39 d0 75 ef 48 8b 05 a8 5b da 00 31 ff c7 44 24 58 00 00 00 00 <48> 89 0f 48 8b 00 48 89 44 24 70 b8 02 00 00 00 66 89 47 1e 89 e8

JerrySievert:
do you have any more info or a reproducible test case?

I still have this active from another issue:

```
CREATE FUNCTION get_env(variable TEXT) RETURNS TEXT AS                                                                         $$
  switch (variable) {
    case 'current_user':  {
      return plv8.execute('SELECT current_user AS val')[0].val;
    }
  }
$$ LANGUAGE plv8;
```

```
do $$ plv8.elog(NOTICE, plv8.find_function('get_env')('current_user')); $$ language plv8;
```

which works exactly as expected.

just also making sure you compiled it with its own v8, not with something from the operating system?

nonara:
Would be glad to provide more detail. Any suggestions on where to look?

I created a new database and added plv8 with your exact function and do command, but I encountered the same issue.

As for v8, it's being built during the make process for plv8. There isn't another on installed on the machine, so I assume it's using the correct version.

Barring isolating the issue further, the steps I believe should reproduce would be to recreate the environment, installing PostgreSQL 12 via yum on CentOS 7.7 - Not sure if that's an option for you.

JerrySievert:
usually steps to reproduce help quite a lot.

nonara:
[...]

PR #375, opened 2019-12-16
Allocate memory for fake_fcinfo

jeffcoat:
Only applies to PG12+.

I can reproduce the crash from https://github.com/plv8/plv8/issues/373 and https://github.com/plv8/plv8/issues/370. This patch fixes it.

More precisely, the change to plv8.cc fixed all the problems that I can reproduce; I made the same change to plv8_func.cc because they're so structurally similar.

### DIFF
--- a/expected/plv8_2.out
+++ b/expected/plv8_2.out
@@ -1,0 +1,740 @@
+-- CREATE FUNCTION
+CREATE FUNCTION plv8_test(keys text[], vals text[]) RETURNS text AS
+$$
+	var o = {};
+	for (var i = 0; i < keys.length; i++)
+		o[keys[i]] = vals[i];
+	return JSON.stringify(o);
+$$
+LANGUAGE plv8 IMMUTABLE STRICT;
+SELECT plv8_test(ARRAY['name', 'age'], ARRAY['Tom', '29']);
+         plv8_test         
+---------------------------
+ {"name":"Tom","age":"29"}
+(1 row)
+
+CREATE FUNCTION unnamed_args(text[], text[]) RETURNS text[] AS
+$$
+	var array1 = arguments[0];
+	var array2 = $2;
+	return array1.concat(array2);
+$$
+LANGUAGE plv8 IMMUTABLE STRICT;
+SELECT unnamed_args(ARRAY['A', 'B'], ARRAY['C', 'D']);
+ unnamed_args 
+--------------
+ {A,B,C,D}
+(1 row)
+
+CREATE FUNCTION concat_strings(VARIADIC args text[]) RETURNS text AS
+$$
+	var result = "";
+	for (var i = 0; i < args.length; i++)
+		if (args[i] != null)
+			result += args[i];
+	return result;
+$$
+LANGUAGE plv8 IMMUTABLE STRICT;
+SELECT concat_strings('A', 'B', NULL, 'C');
+ concat_strings 
+----------------
+ ABC
+(1 row)
+
+CREATE FUNCTION return_void() RETURNS void AS $$ $$ LANGUAGE plv8;
+SELECT return_void();
+ return_void 
+-------------
+ 
+(1 row)
+
+CREATE FUNCTION return_null() RETURNS text AS $$ return null; $$ LANGUAGE plv8;
+SELECT r, r IS NULL AS isnull FROM return_null() AS r;
+ r | isnull 
+---+--------
+   | t
+(1 row)
+
+-- TYPE CONVERTIONS
+CREATE FUNCTION int2_to_int4(x int2) RETURNS int4 AS $$ return x; $$ LANGUAGE plv8;
+SELECT int2_to_int4(24::int2);
+ int2_to_int4 
+--------------
+           24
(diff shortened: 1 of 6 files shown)


JerrySievert:
awesome! I'll wait to see what CI looks like before merging.

jeffcoat:
I'm very new to this code and PG extensions and don't really know what I'm doing. I'm mostly hoping this will point someone who does in the right direction.

JerrySievert:
hm. reviewing this, it doesn't actually do what you might think it does, and also causes a memory leak.

I'll comment in-code.

plv8.cc
```diff
@@ -1620,7 +1620,8 @@ GetGlobalContext(Persistent<Context>& global_context)
 				fake_fcinfo.arg[1] = CStringGetDatum(arg);
 				Datum ret = has_function_privilege_id(&fake_fcinfo);
 #else
-				MemSet(fake_fcinfo, 0, sizeof(fake_fcinfo));
+				fake_fcinfo = (FunctionCallInfo) palloc(SizeForFunctionCallInfo(2));
```
JerrySievert: here you're allocating over an existing object.

JerrySievert: also missing a `pfree()` so causing a memory leak

jeffcoat: I can believe that. (Though I don't see how. There's a declaration `FunctionCallInfo fake_fcinfo;` on line 1601, then this is the first statement that refers to fake_fcinfo following the declaration.)

Without the allocation, I get a segfault at the line `MemSet(fake_fcinfo, 0, sizeof(fake_fcinfo));` I'm interpreting that as an attempt to zero out unallocated memory.

Backtrace from a core file:

```
Core was generated by `postgres: 12/main: import import ::1(42918) SELECT                            '.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fdeeb657a2c in GetGlobalContext (global_context=...) at plv8.cc:1625
1625					fake_fcinfo->flinfo = &flinfo;
[Current thread is 1 (Thread 0x7fdf710b8a40 (LWP 1747400))]
(gdb) bt
#0  0x00007fdeeb657a2c in GetGlobalContext (global_context=...) at plv8.cc:1625
```

Which shows in syslog as
```
Dec 16 17:01:39 behenian kernel: [1559398.124502] postgres[1747400]: segfault at 0 ip 00007fdeeb657a2c sp 00007ffc33f6d320 error 6 in plv8-3.0alpha.so[7fdeeb64a000+934000]
Dec 16 17:01:39 behenian kernel: [1559398.124517] Code: 1f 80 00 00 00 00 48 83 c0 08 48 c7 40 f8 00 00 00 00 48 39 d0 75 ef 48 8b 05 c8 e3 b9 00 31 ff c7 85 d8 fe ff ff 00 00 00 00 <48> 89 0f 48 8b 00 48 89 85 f0 fe ff ff b8 02 00 00 00 66 89 47 1e
```

JerrySievert: likely what occurred is that Postgres internals changed the `typedef` for FunctionCallInfo.  I'll take a deeper look into it when I get the time.

JerrySievert: if that's the case, then it will need to be `#if`'d to work correctly across all postgres versions that plv8 supports.

jeffcoat: Thank you for your feedback. 

* You're right about the changed typedef for FunctionCallInfo. You set up the  `#if`s a few months ago in commit 77b941128cdb; I'm building off that.

* To avoid memory leaks, I've changed fake_fcinfo to be stack-allocated, instead of using palloc(). The LOCAL_FCINFO macro is from the PG12 version of fmgr.h.

jeffcoat:
I don't think the CI build failure was caused by a problem on this branch, but I don't have the authority to re-try it.

JerrySievert:
I restarted the job

jeffcoat:
Thank you; much better.

expected/plv8_2.out
```diff
@@ -0,0 +1,740 @@
+-- CREATE FUNCTION
```
JerrySievert, line 1: what are you trying to add here?

jeffcoat: Sorry; it's explained in the commit message, but this UI doesn't show that unless you go looking for it.

One of the tests in plv8.sql -- `text_to_int4('abc')` -- fails under PG12 because the error message changed from `invalid input syntax for integer:` to `invalid input syntax for type integer:`.

I was trying to make the tests pass on all versions with minimal re-writing, so I added plv8_2.out as an alternative expected/ file that uses the new error message. I'd be happy to change it if you'd prefer split this case into a separate test file, or something like that.

JerrySievert: I think I'd prefer a second PR, separate out the concerns.

jeffcoat: No problem; I put them together here because the tests won't pass against PG12 without the fcinfo change. Which speaks well for the test suite.

JerrySievert: updating `.travis.yml` should also go in the other PR :)

PR #376, opened 2019-12-18
Make regression tests PG12 compatible

jeffcoat:
No functional changes, just matching output formats.

* plv8_2.out is a copy of plv8.out except an error message changes from `invalid input syntax for integer:` to `invalid input syntax for type integer:`.
* The default value of `extra_float_digits` changed, so set it explicitly in window.sql so it'll work either way

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_script:
 
 env:
   matrix:
+    - PGVERSION=12
     - PGVERSION=11
     - PGVERSION=10
     - PGVERSION=9.6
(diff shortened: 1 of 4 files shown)


JerrySievert:
can you add your `.travis.yml` changes here as well?  after that I'll go ahead and merge after CI passes.

thanks!

jeffcoat:
Added PGVERSION=12, with the expectation that that environment will fail CI without the second PR.

JerrySievert:
ok, I'll go ahead and merge this - when all is done I'll go ahead and back port everything to the 2.3 branch.

Issue #393, opened 2020-06-16
user context dies in pretty strange circumstances


pkit:
Yup. I see something like this in log:
```
2020-06-14 13:53:24.753 UTC [82/44290] 5ee49a79.52 ERROR:  missing error text
2020-06-14 13:53:24.753 UTC [82/44291] 5ee49a79.52 CONTEXT:
```
And the user context is dead on that connection: all the globals are deleted, `plv8_start` is not run, looks like a context was restarted(?) for some reason, but none of the bootstrap was run.
Not connected to any specific query or javascript function (although does happen much more frequently on frequenlty launched functions) and is overall pretty rare.
But still annoying because there is no "bad" condition that I can anchor it to.
Need to rely on heuristics: is the context really alive or not?
Needless to say: all other contexts on the same backend continue to function properly.
If there are any pointers on how to debug that one, will be appreciated.

pkit:
Looks like `GetGlobalContext()` throws something...
But it's so hard to reproduce...

JerrySievert:
hm.  if anyone could figure out how to reproduce in the wild, I'd guess it was you. :)

so I can think of better ways to debug, which version are you using in this case? 2.3 or 3.0?

JerrySievert:
also knowing the pg version might help, trying to track down exactly where "missing error text" might come from.

pkit:
I'm using the HEAD which is `r3.0alpha`
With PG 10.10
Did a workaround: created a stored proc to reset the user context properly `plv8_reset()` for now, if you want it in a PR, can do (I reset the context only if I see it's busted, much faster than reconnect, and doesn't interfere with the other contexts)
Do you have any pointers on why the error text can be missing?
Have another wild guess: all the functions that lead to the error use txid which is bigint as the last argument. May be a pure coincidence.

JerrySievert:
first off, bear with me - slowly getting my context switched.  while I'm typically deep in c++ every day, it's been in DSP, not DB, so need to swap that portion of my brain back in :)

> if you want it in a PR, can do (I reset the context only if I see it's busted, much faster than reconnect, and doesn't interfere with the other contexts)

yeah, I'd definitely accept a PR for this - even when the issue is fixed, it will be nice to have the *option* of reseting the context.

> Do you have any pointers on why the error text can be missing?

that's what I'm unsure about, and why I asked about the pg version - I'll spend a few minutes trying to track down that portion of it.

[...]

PR #395, opened 2020-06-18
plv8_reset() to reset context

pkit:
adds `plv8_reset()` that resets the context and re-runs init: `plv8_start` proc when needed
see #393


JerrySievert:
hm.  not sure why it thinks that travis hasn't finished yet, but it all appears to pass.  merging this first then will wait on the updated travis on the second pr.

thanks again!

PR #397, opened 2020-06-26
Changes to build against PG13. This includes the removal of a referen…

jim-mlodgenski:
…ce to OPAQUEOID which has been deprecated long before PLV8 was invented.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -38,6 +38,10 @@ extern "C" {
 #include "catalog/pg_database.h"
 #endif
 
+#if PG_VERSION_NUM >= 130000
+#include "common/hashfn.h"
+#endif
+
 #include <signal.h>
 
 #ifdef EXECUTION_TIMEOUT
@@ -991,9 +995,7 @@ common_pl_call_validator(PG_FUNCTION_ARGS, Dialect dialect) throw()
 	/* except for TRIGGER, RECORD, INTERNAL, VOID or polymorphic types */
 	if (functyptype == TYPTYPE_PSEUDO)
 	{
-		/* we assume OPAQUE with no arguments means a trigger */
-		if (proc->prorettype == TRIGGEROID ||
-			(proc->prorettype == OPAQUEOID && proc->pronargs == 0))
+                if (proc->prorettype == TRIGGEROID)
 			is_trigger = true;
 		else if (proc->prorettype != RECORDOID &&
 			proc->prorettype != VOIDOID &&
--- a/plv8.cc
+++ b/plv8.cc
@@ -995,7 +995,13 @@ common_pl_call_validator(PG_FUNCTION_ARGS, Dialect dialect) throw()
 	/* except for TRIGGER, RECORD, INTERNAL, VOID or polymorphic types */
 	if (functyptype == TYPTYPE_PSEUDO)
 	{
+#if PG_VERSION_NUM >= 130000
                 if (proc->prorettype == TRIGGEROID)
+#else
+                /* we assume OPAQUE with no arguments means a trigger */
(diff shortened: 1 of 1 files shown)


JerrySievert:
when did it become deprecated?  from conversations while plv8 was being written in 2011, it appears that it was still a valid way to check for a trigger.

is there a version cutoff where this behavior changed?

jim-mlodgenski:
On Fri, Jun 26, 2020 at 10:51 AM Jerry Sievert <notifications@github.com>
wrote:

> when did it become deprecated? from conversations while plv8 was being
> written in 2011, it appears that it was still a valid way to check for a
> trigger.
>
It was there to support functions created back in 7.3 and before.

> is there a version cutoff where this behavior changed?
>
Anything created after 8.2 would never have OPAQUEOID as a return type
https://www.postgresql.org/message-id/4110.1583255415%40sss.pgh.pa.us

JerrySievert:
> Anything created after 8.2 would never have OPAQUEOID as a return type

that's about what I thought.  I went and checked my notes, and supposedly we're still able to support all the way down to 8.1 (I haven't tried to compile against it in a while).

let me give this some thought, as to whether to deprecate anything older than, say, 9.1 for this version.

jim-mlodgenski:
On Fri, Jun 26, 2020 at 11:09 AM Jerry Sievert <notifications@github.com>
wrote:

> Anything created after 8.2 would never have OPAQUEOID as a return type
>
> that's about what I thought. I went and checked my notes, and supposedly
> we're still able to support all the way down to 8.1 (I haven't tried to
> compile against it in a while).
>
> let me give this some thought, as to whether to deprecate anything older
> than, say, 9.1 for this version.
>
The main PG community only supports 9.5 on up at this point with 9.5 being
deprecated in 8 months. Anyone still running 9.1 or older has been running
on an unsupported version for a very long time.

JerrySievert:
> The main PG community only supports 9.5 on up at this point with 9.5 being
deprecated in 8 months. Anyone still running 9.1 or older has been running
on an unsupported version for a very long time.

yes, I realize that, but plv8 has a different deprecation schedule, because of a set of users.

not to mention that I still see posts on pg-bugs on a fairly regular basis with people still stuck using older versions of pg.

those unsupported versions can still be very hard to move off of, especially in a larger group that is risk adverse.

jim-mlodgenski:
On Fri, Jun 26, 2020 at 11:20 AM Jerry Sievert <notifications@github.com>
wrote:

> The main PG community only supports 9.5 on up at this point with 9.5 being
> deprecated in 8 months. Anyone still running 9.1 or older has been running
> on an unsupported version for a very long time.
>
> yes, I realize that, but plv8 has a different deprecation schedule,
> because of a set of users.
>
> not to mention that I still see posts on pg-bugs on a fairly regular basis
> with people still stuck using older versions of pg.
>
> those unsupported versions can still be very hard to move off of,
> especially in a larger group that is risk adverse.
>

To be conservative then, we can just put an "if PG_VERSION_NUM >= 130000"
around the OPAQUEOID change and things would continue to work the same way
on older versions.

JerrySievert:
that would be a fantastic compromise.  would you be willing to make that change on your PR?

jim-mlodgenski:
On Fri, Jun 26, 2020 at 11:24 AM Jerry Sievert <notifications@github.com>
wrote:

> that would be a fantastic compromise. would you be willing to make that
> change on your PR?
>

No problem. I made the change.

JerrySievert:
awesome, thanks very much for the PR!  I'll let the CI run and will plan to do a merge this afternoon (pacific time).

I can do the back port to the 2.x branch.

PR #399, opened 2020-06-30
use isolate per user

pkit:
creates isolate for each user
make full per user accounting possible even on the same
connection/backend
reduces accidental damage from badly written javascript code: user
isolates can be dropped while retaining all other users


pkit:
Dunno if you want it. But just in case. :)
Using it in my project.

JerrySievert:
woo! you had mentioned it a while back, was hoping I'd get to see it.

do you know how much of a performance/memory impact it makes, if any?

JerrySievert:
(this is going to take me a few days to go through thoroughly, btw)

pkit:
Didn't measure it specifically. But the obvious things are:
1. It will use the initial snapshot per user now, which is at least 4MB AFAIR (but I'm using a pretty heavy start function which eats up at least 16MB).
2. Static variables cannot be reused anymore, means that all the global, window, cursor, etc. prototypes are per-isolate now.
3. Will check its perf soon enough though :)

pkit:
> (this is going to take me a few days to go through thoroughly, btw)

Yup. It's a pretty big change...
And some code is kinda questionable, but I was trying to compromise between clarity and the tight coupling...
`current_context` global is pretty finicky because of the order.

JerrySievert:
it looked pretty good - a couple of things caught my eye, but nothing too major.

PR #405, opened 2020-09-11
fixed the PostgreSQL error "attempt to redefine parameter"

mpietrobo:
This error is caused by trying to redeclare a GUC variable that already exists.
It may happen during an upgrade of PLV8 if the previous library is loaded before the new one.
In our case it happened when upgrading from 2.3.13 to 2.3.15.
Since it seems that PostgreSQL API is not publishing a function to check whether a GUC variable already exists or not, this patch used the approach of scanning through all declared variables and check whether the ones from PLV8 are already in that list or not. If a variable is in the list then it reads the current value, if not then it is created.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -43,6 +43,8 @@ extern "C" {
 #endif
 #endif
 
+#include <stdlib.h>
+
 PG_MODULE_MAGIC;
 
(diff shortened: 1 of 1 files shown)


JerrySievert:
the tests are showing that not only are GUC's not being properly defined, but there is also some corruption going on.

so, it looks like this approach isn't going to work as-is.

PR #406, opened 2020-09-11
fixed the PostgreSQL error "attempt to redefine parameter"

mpietrobo:
This error is caused by trying to redeclare a GUC variable that already exists.
It may happen during an upgrade of PLV8 if the previous library is loaded before the new one.
In our case it happened when upgrading from 2.3.13 to 2.3.15.
This is basically the same as the other fix but for the branch r3.0alpha.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -52,6 +52,8 @@ extern "C" {
 #endif
 #endif
 
+#include <stdlib.h>
+
 PG_MODULE_MAGIC;
 
 PGDLLEXPORT Datum	plv8_call_handler(PG_FUNCTION_ARGS);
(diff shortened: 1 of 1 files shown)


mpietrobo:
sorry, I've found a way to fix the message "WARNING: unrecognized configuration parameter "plv8.start_proc"" but there is still an issue with "RESET plv8.start_proc" that is somehow causing that parameter to be set to some invalid address instead of NULL, and it seems to be in PostgreSQL code that is not exported via guc.h so I cannot access and fix it.
I'm trying to find a workaround but it's complicated, because we have no access to the real union and structs used to store the declaration of the GUC variable in guc.cc.

JerrySievert:
ok, I haven't had time to do a full code review (sorry), but I'm fairly adept at traversing pg source code at this point, so happy to spend some time tomorrow taking a look.

mpietrobo:
well, the patch is really simple: it cycles through all declared GUC variables using GetNumConfigOptions and GetConfigOptionByNum from guc.h and checks if any of them matches one of the options that we want to create. If so, it reads the current value to set the corresponding static variable.
All plv8's options that are still marked as non existant at the end of the first for loop will be explicitly declared in the second loop using DefineCustomIntVariable or DefineCustomStringVariable.
To simplify the code, the plv8 's variables that we want to add are stored in a small static array, to avoid having to fill the loop with a cascade of it/then/elseif.

pkit:
@mpietrobo probably because it's a `GUC_CUSTOM_PLACEHOLDER ` and not real var
You need to check for that and re-declare (re-declaring over placeholder is safe, AFAIR)
See here the official handling in guc.c https://github.com/postgres/postgres/blob/3c8819955023694feeaa456ee60853d0d6d0e60a/src/backend/utils/misc/guc.c#L5249

pkit (commented):
Seems over-complicated to me. Just a simple if/else per varable would be much more readable.
```
#define VAR_NAME "plv8.var_name"
if (plv8_guc_variable_exists(VAR_NAME )) {
  elog(....something, VAR_NAME);
} else {
  DeclareStringVariable(..., VAR_NAME ,....);
}
```

pkit:
@mpietrobo see #409 
Was a pretty rough ride

PR #407, opened 2020-09-26
fix plv8_reset isolate disposal

pkit:
some of my assumption on how stuff is disposed off were incorrect
fixed it

`Isolate::Dispose()` uses pretty complex logic that deletes stuff in particular order
And the main problem: it deletes the array allocator, which means managing allocator lifecycle externally is kinda prohibited
Got SEGFAULTs in pretty convoluted cases, which are hard to reproduce with a simple test...


pkit:
I would mark it as WIP though for now. Because may find other stuff there.

pkit:
close in favor of #441

PR #412, opened 2020-11-03
fixed bigint's bug

qorm:


JerrySievert:
do you have a test that shows why this is needed?

qorm:
if column type is bigint and this number > int32. JSON.stringify don't work

JerrySievert:
`json.stringify()` explicitly does not work, by design.

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Usage_recommendations for more details.

Issue #413, opened 2020-11-10
Building for ARM


inian:
Hi, I am trying to get plv8 to build on ARM. With some changes, I was able to get v8 to build on ARM. `make` and `make install` seems to work too. But when I enable the extension on database, I get the following error. 

```
Query 1 ERROR: ERROR:  could not load library "/usr/lib/postgresql/12/lib/plv8-3.0alpha.so": /usr/lib/postgresql/12/lib/plv8-3.0alpha.so: undefined symbol: _ZN6icu_638ByteSinkD2Ev
```

Any help on getting this to work would be appreciated! With ARM growing in popularity, having official build instructions for ARM would help a lot.

JerrySievert:
`c++filt` tells me that the undefined symbol is `icu_63::ByteSink::~ByteSink()`

it's possible that you have a different version of icu installed than what v8 is being built with, or you might try disabling icu for the v8 build.  can you expand on which arm platform and which arm os you're using?

inian:
I am trying to build it with a AWS Graviton instance running Ubuntu 18.04.
```
> uname -a
Linux ip-xx 5.4.0-1029-aws #30~18.04.1-Ubuntu SMP Tue Oct 20 11:08:38 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

ICU support is [disabled by default](https://github.com/plv8/plv8/blob/650340187526aaf7160e62973fa930954379214b/Makefile#L22) right?

JerrySievert:
it is disabled by default, but v8 is apparently trying to link it in regardless.

I can set up a pi 3b and try to build, but do not currently have any graviton instances available to me.

inian:
Hey thanks for the tip, I got it working now :) Turns out I hadn't disabled ICU support properly. 

I can submit a patch for making plv8 work on ARM if you are interested, but it pretty brittle. I used [this code](https://github.com/sqreen/PyMiniRacer/blob/4e4d0731a74e6f7a0a55186fb5078ae90b06e19a/build_linux_arm64.sh) here as a starting point to get v8 to compile on ARM.

JerrySievert:
hm. 3b is 32bit.  I can only advise, not attempt the build myself.

JerrySievert:
sure, I'd love to see a PR - there is already some code for detecting arm that could be augmented.

PR #414, opened 2020-11-11
adds arm64 support

inian:
Hi as I mentioned [here](https://github.com/plv8/plv8/issues/413#issuecomment-724826905), I have got support for arm64. Tested it on a AWS Graviton instance running ubuntu18.04. 

The installation process for v8 on ARM is not pretty though. I built upon the code from [here](https://github.com/sqreen/PyMiniRacer/blob/4e4d0731a74e6f7a0a55186fb5078ae90b06e19a/build_linux_arm64.sh) and [here](https://github.com/hypersleep/rails-on-arm/blob/master/Dockerfile).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,31 @@
 # structure in v8 which may be different from version to another, but user
 # can specify the v8 version by AUTOV8_VERSION, too.
 #-----------------------------------------------------------------------------#
+
+# Platform detection
+ifeq ($(OS),Windows_NT)
+	# noop for now
+else
+	UNAME_S := $(shell uname -s)
+	ifeq ($(UNAME_S),Darwin)
+		PLATFORM = x64.release
+	endif
+	ifeq ($(UNAME_S),Linux)
+		ifeq ($(shell uname -m | grep -o arm),arm)
+			PLATFORM = arm64.release
+		endif
+		ifeq ($(shell uname -m),aarch64)
+			PLATFORM = arm64.release
+		endif
+		ifeq ($(shell uname -m),x86_64)
+			PLATFORM = x64.release
+		endif
+	endif
+endif
+
 AUTOV8_VERSION = 7.4.288.28
 AUTOV8_DIR = build/v8
-AUTOV8_OUT = build/v8/out.gn/x64.release/obj
+AUTOV8_OUT = build/v8/out.gn/$(PLATFORM)/obj
 AUTOV8_DEPOT_TOOLS = build/depot_tools
 AUTOV8_LIB = $(AUTOV8_OUT)/libv8_snapshot.a
 AUTOV8_STATIC_LIBS = -lv8_base -lv8_snapshot -lv8_libplatform -lv8_libbase -lv8_libsampler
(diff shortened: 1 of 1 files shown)


JerrySievert:
awesome.  but given that you're downloading stuff, I'm going to need to find an arm64 system to test on before I can accept the PR.

inian:
Cool. Let me know if you need a graviton instance to test this out on.

JerrySievert:
that would be quite helpful.


public ssh key:

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC/tdD4TEHbcoJeGsSifP58vXtfWk71Gt43+0oQrSd32jHy4XPldjvRyQxHxz583EKFBS5IcreOESgD6LN1JOE9/cJj5/Xcemv23Y9jMUKNyHVcTaCjzk6oLE4EOuN/8dyUzSBJXdj2Rvkm8a44v9RmzyGkfhVzCN9Pp8H+TfA0SHtibasthl1WcKGU/NHN/GL7frBTwQgba22hWgAgGFAVAUJjJ8+Kwa+X0sqKqiO1Tg6XpA5GhH0xSTXgzFJWooccZcL2JVWfPY+9NQYNsZPvR+pNc/khzqrJg2wgEamMiv/BAhZ67M8ESnrtfA3/RhGx6sJwjG5cKBNzZrs2ys4dVVk5q+DyALmRq+q99IAbT6LDSqXvzGSaD6PlAb5LxFTb043yxPH06E0HjWxChrJ0pwO8ZxypXO9Xau/Mgd+KgkCfHO2etkx3U3oV+MT/uMM4Qh4Uo8HIi6EHOT/f9hJz9D9AbQw+LOD3NDDBsuh0UtOHxLLpcno9Z12UYgbdD8TAF5QQgDUS4GiTAyScWv+zC6ADjn5koc13JJ6+2k0CzAnxpeozXAUcs8w5mnEiW/HXSq7a0BM3RP8gkRvlXA91O1P5RHOSrsQrgG7cG5+a9HVZkSzXK/dKBPt+AcPw3DLjh2RJE9/oL0zDclPch7Y3IL7uC+iqPVDE5DFIUx1Nfw== code@legitimatesounding.com
```

inian:
hey set up an arm instance running ubuntu 18.04 at ubuntu@54.250.55.39

JerrySievert:
awesome, thanks, I'm in!

JerrySievert:
some notes so far:

* requires ninja-build package to be installed
* /bin/sh: 1: clang++: not found at build

and then the build fails for me - do you have a list of dependencies you installed?

Makefile
```diff
@@ -31,8 +53,33 @@ $(AUTOV8_DEPOT_TOOLS):
...
+ifeq ($(PLATFORM),arm64.release)
+$(AUTOV8_DIR): $(AUTOV8_DEPOT_TOOLS)
+	# patch with system-installed ninja
+	cp /usr/bin/ninja build/depot_tools/
+
+	# get ARM64 clang binaries
+	cd build; wget -nc http://releases.llvm.org/7.0.1/clang+llvm-7.0.1-aarch64-linux-gnu.tar.xz; tar xf clang+llvm-7.0.1-aarch64-linux-gnu.tar.xz
+
+	# Get and build the plugin
+	cd build; wget -nc https://chromium.googlesource.com/chromium/src/+archive/lkgr/tools/clang/plugins.tar.gz; mkdir -p plugin; cd plugin; tar xf ../plugins.tar.gz; clang++ *.cpp -c -I ../clang+llvm-7.0.1-aarch64-linux-gnu/include/ -fPIC -Wall -std=c++14 -fno-rtti -fno-omit-frame-pointer; clang -shared *.o -o libFindBadConstructs.so
```
JerrySievert: `clang++` should be changed to `../clang+llvm-7.0.1-aarch64-linux-gnu/bin/clang++`

`clang` should be changed to `../clang+llvm-7.0.1-aarch64-linux-gnu/bin/clang`

or a path needs to be set that includes the directory.

`clang++` also does not appear to be in the path for the actual build of v8, so it looks like you have another dependency installed that isn't documented.

Makefile
```diff
@@ -31,8 +53,33 @@ $(AUTOV8_DEPOT_TOOLS):
...
+
+	cp build/plugin/libFindBadConstructs.so build/clang+llvm-7.0.1-aarch64-linux-gnu/lib/
+
+	# Build an ARM64 binary of gn
+	cd build; rm -rf gn; git clone https://gn.googlesource.com/gn; cd gn; git checkout 6ae63300be3e9865a72772e4cb6e1f8f667624c4; sed -i -e "s/-Wl,--icf=all//" build/gen.py; python build/gen.py; ninja -C out
```
JerrySievert: looks like you should change this line to:

```
	cd build; rm -rf gn; git clone https://gn.googlesource.com/gn; cd gn; git checkout 6ae63300be3e9865a72772e4cb6e1f8f667624c4; sed -i -e "s/-Wl,--icf=all//" build/gen.py; python build/gen.py; export PATH=$$PATH:`pwd`/../clang+llvm-7.0.1-aarch64-linux-gnu/bin ; ninja -C out
```

JerrySievert:
ok, see my notes - after those changes it looks like all tests pass.  can you make those changes to your PR and then I'll test one more time?

thanks!

inian:
Hi Jerry, yes I forgot to mention that these 2 additional dependencies need to be installed at the system level 
> apt install clang ninja-build

Here are the complete list of dependencies I used to get it working

      - build-essential
      - ca-certificates
      - curl
      - git-core
      - gpp
      - cpp
      - pkg-config
      - apt-transport-https
      - cmake
      - libc++-dev
      - libc++abi-dev
      - postgresql-server-dev-12
      - libc++1
      - libtinfo5
      - libc++abi1
      - clang
      - ninja-build

Makefile
inian: The clang package would be a system-level dependency. Once this is installed, clang and clang++ would be in the path.

Makefile
inian: This change would also not be required if clang is installed as a system dependency.

JerrySievert:
with my changes, the only additional dependency  beyond what's already documented is `ninja-build`.  since you're already downloading `llvm`, I'd rather that version get used in the build.

inian:
Okay agreed. I have pushed a commit to add the local llvm to the path and doesn't rely on a system dependency.

JerrySievert:
awesome, looks much better!  thanks much!

also, thanks for the aws instance, I'm now done with it!

inian:
Thanks for testing it and merging it in Jerry!

JerrySievert:
I thought you might be interested to know that I managed to get a pi running 64bit ubuntu going, and it compiles great with one small change!

thanks again!

PR #424, opened 2021-01-10
bug: string input in bytea garbled

pkit:
fixes a garbage created in byte fields if string is used as input and
not an *Array type


pkit:
Probably should be back-ported to 2.3.x too :)

JerrySievert:
nice catch!  I wonder how long that's been missing.

and yeah, I'll go ahead and back port that :)

PR #426, opened 2021-01-23
Improve ArrayAllocator

pkit:
- ArrayAllocator::Reallocate added
- streamlined some of the heap usage math


pkit:
Ok, the `Reallocate` is `final` in v8 < 8.x
Thus this one need to go in after 8.8 upgrade...

PR #428, opened 2021-01-23
Upgrade to v8 8.8 and improve some features

pkit:
- upgrade to v8 8.8.278.14
- improve timeouts, add timeout to script compilation
- improve OOM handling
- add fixes for common exploits


pkit:
Travis ci time limit exceeded building v8...
I suppose migrating to Github Actions is one option

P.S. the funny thing is that docker builds are correctly parallelized in 8.8, it builds almost 3x faster on my laptop.
But Travis CI instances are tiny 1 CPU...

pkit:
Yup, seems like github actions is a way to go. They allow for 6 hour builds.
I'm also using staging images internally to not rebuild v8 if it's not needed (same v8 version, same patches)
But they are pretty tightly integrated in my local build process will be hard to isolate...

JerrySievert:
yeah.  I was an original backer of travis, so I've taken my time to move away.  with the purchase, it's been more pressing, but with the purchase of GitHub by Microsoft it's felt like being between a rock and a hard place.  I'll investigate, especially given the new amount of time that v8 takes to build, but no free options are really feeling good at this time.  I may just need to bite the bullet and pay for a service.

Makefile
```diff
@@ -29,16 +29,18 @@ else
...
 endif
 
-AUTOV8_VERSION = 7.4.288.28
+AUTOV8_VERSION = 8.8.278.14
 AUTOV8_DIR = build/v8
 AUTOV8_OUT = build/v8/out.gn/$(PLATFORM)/obj
 AUTOV8_DEPOT_TOOLS = build/depot_tools
 AUTOV8_LIB = $(AUTOV8_OUT)/libv8_snapshot.a
-AUTOV8_STATIC_LIBS = -lv8_base -lv8_snapshot -lv8_libplatform -lv8_libbase -lv8_libsampler
+AUTOV8_STATIC_LIBS = -lv8_base_without_compiler -lv8_compiler -lv8_base_without_compiler -lchrome_zlib -lcompression_utils_portable -ltorque_generated_definitions -lcppgc_base -lv8_cppgc_shared -lv8_snapshot -lv8_libplatform -lv8_libbase -lv8_libsampler -lchrome_zlib
```
JerrySievert: why not just use `v8_monolith` here?

Makefile
```diff
@@ -29,16 +29,18 @@ else
...
+ZLIB_DIR = $(AUTOV8_OUT)/third_party/zlib
+ZLIB_OBJS = $(ZLIB_DIR)/zlib_adler32_simd/adler32_simd.o $(ZLIB_DIR)/zlib_inflate_chunk_simd/inffast_chunk.o $(ZLIB_DIR)/zlib_inflate_chunk_simd/inflate.o $(ZLIB_DIR)/zlib_crc32_simd/crc32_simd.o $(ZLIB_DIR)/zlib_x86_simd/crc_folding.o $(ZLIB_DIR)/zlib_x86_simd/fill_window_sse.o
 export PATH := $(abspath $(AUTOV8_DEPOT_TOOLS)):$(PATH)
 
-SHLIB_LINK += -L$(AUTOV8_OUT) -L$(AUTOV8_OUT)/third_party/icu $(AUTOV8_STATIC_LIBS)
-V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_snapshot=true v8_use_external_startup_data=false use_custom_libcxx=false
+SHLIB_LINK += $(ZLIB_OBJS) -L$(AUTOV8_OUT) -L$(AUTOV8_OUT)/third_party/icu -L$(ZLIB_DIR) -L$(AUTOV8_OUT)/third_party/zlib/google $(AUTOV8_STATIC_LIBS)
+V8_OPTIONS = is_component_build=false v8_static_library=true v8_use_external_startup_data=false use_custom_libcxx=false
```
JerrySievert: adding `v8_monolithic=true` defaults a lot of this

Makefile
```diff
@@ -101,6 +112,9 @@ endif
 # enable direct jsonb conversion by default
 CCFLAGS += -DJSONB_DIRECT_CONVERSION
 
+# enable pointer compression support (default for v8 > 8.x)
+CCFLAGS += -DV8_COMPRESS_POINTERS -DV8_31BIT_SMIS_ON_64BIT_ARCH
+
```
JerrySievert, line 149: yup, hit this one a few times :)

plv8.cc
```diff
@@ -1862,26 +1944,39 @@ GetGlobalObjectTemplate(Isolate *isolate)
...
+		templ->Set(String::NewFromUtf8(isolate, "LOG", NewStringType::kInternalized).ToLocalChecked(),
+				   Int32::New(isolate, LOG));
+		templ->Set(String::NewFromUtf8(isolate, "INFO", NewStringType::kInternalized).ToLocalChecked(),
+				   Int32::New(isolate, INFO));
+		templ->Set(String::NewFromUtf8(isolate, "NOTICE", NewStringType::kInternalized).ToLocalChecked(),
+				   Int32::New(isolate, NOTICE));
+		templ->Set(String::NewFromUtf8(isolate, "WARNING", NewStringType::kInternalized).ToLocalChecked(),
+				   Int32::New(isolate, WARNING));
+		templ->Set(String::NewFromUtf8(isolate, "ERROR", NewStringType::kInternalized).ToLocalChecked(),
+				   Int32::New(isolate, ERROR));
```
JerrySievert, line 2045: I think the 1000's of changes to `ToLocalChecked()` were the most frustrating to me - see my personal fork.

plv8.cc
```diff
@@ -2019,6 +2115,7 @@ Datum
 Converter::ToDatum(Handle<v8::Value> value, Tuplestorestate *tupstore)
 {
 	Isolate		   *isolate = Isolate::GetCurrent();
+    Local<Context>  context = isolate->GetCurrentContext();
```
JerrySievert, line 2195: and all of the hundreds of places where local context is used gratuitously, which is rather frustrating.

plv8_func.cc
```diff
@@ -102,7 +102,7 @@ SetCallback(Handle<ObjectTemplate> obj, const char *name,
 			FunctionCallback func, PropertyAttribute attr = None)
 {
 	Isolate* isolate = Isolate::GetCurrent();
-	obj->Set(String::NewFromUtf8(isolate, name, String::kInternalizedString),
+	obj->Set(String::NewFromUtf8(isolate, name, NewStringType::kInternalized).ToLocalChecked(),
```
JerrySievert, line 105: I don't think we need to actually use `NewStringType` in these places, since it gracefully degrades.

plv8_func.cc
```diff
@@ -376,19 +382,20 @@ plv8_FunctionInvoker(const FunctionCallbackInfo<v8::Value> &args) throw()
...
+        err->Set(context, String::NewFromUtf8(isolate, "schema_name").ToLocalChecked(), schema_name).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "table_name").ToLocalChecked(), table_name).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "column_name").ToLocalChecked(), column_name).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "datatype_name").ToLocalChecked(), datatype_name).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "constraint_name").ToLocalChecked(), constraint_name).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "detail").ToLocalChecked(), detail).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "hint").ToLocalChecked(), hint).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "context").ToLocalChecked(), err_context).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "internalquery").ToLocalChecked(), internalquery).Check();
+        err->Set(context, String::NewFromUtf8(isolate, "code").ToLocalChecked(), code).Check();
```
JerrySievert: this, I think, is my most frustrating of all - no consistent use of `context`, `isolate`, or lack thereof (not your fault, just my frustration at v8)

plv8_type.cc
```diff
@@ -1305,16 +1315,16 @@ CString::~CString()
...
+	if(!value.IsEmpty()) {
+		Isolate* isolate = Isolate::GetCurrent();
+		Local<String> str;
+		if (value->ToString(isolate->GetCurrentContext()).ToLocal(&str)) {
+			String::Utf8Value utf8(isolate, str);
+			if(*utf8) {
+				out = *utf8;
+				return true;
+			}
+		}
```
JerrySievert, line 1327: yup, see my previous comment about this in another bug.

sql/exploits.sql
```diff
@@ -0,0 +1,36 @@
...
+DO $$
+    [].sort.call('1'.repeat(0x7ffffff));
+$$ language plv8;
+select exploit_alive();
+DO $$
+    JSON.parse('[' + '1,'.repeat(0x7fffffd) + '1]');
+$$ language plv8;
+select exploit_alive();
+
+DROP FUNCTION exploit_alive;
```
JerrySievert, line 36: all of this is awesome.

JerrySievert (commented):
mostly just frustration with v8 :)

Makefile
pkit: Can you list the options? I have failed to produce a build with no UND in symbol table... :)

plv8_func.cc
pkit: Ah, yeah, it's left from earlier versions. In newer ones they even have convenience functions for that.  Need to update...

sql/exploits.sql
pkit: You can paste these into Chrome(ium) and see how it explodes spectacularly!
And all of that happens because somebody thought that MAX_STRING_LENGTH = 2 * MAX_ARRAY_LENGTH would be so much fun, while keeping Array interface in String... :)

Makefile
JerrySievert: `use_custom_libcxx=false v8_monolithic=true v8_use_external_startup_data=false is_component_build=false`

and then build with:

`cd $(AUTOV8_DIR) ; env CXXFLAGS=-fPIC CFLAGS=-fPIC ninja -C out.gn/$(PLATFORM) v8_monolith`

this has worked well for me.

plv8_func.cc
JerrySievert: it's not just a need to update, but in some versions of v8 it actually produces bad strings.  very frustrating.  removing it fixes the bad string creation.  typically, it's something like `data\u37`.  so it appears to be bad unicode conversion in v8 itself.   I believe they've fixed it in more recent versions, but I'd rather not be on the bad side of that when they screw up and revert a patch.

sql/exploits.sql
JerrySievert: I'd rather not :). I compiled and tried a fork bomb on a sparc 5 in the 90's, just to see for myself what it did, but in this case I'll just continue to use safari.

plv8_func.cc
pkit: Do you want to use Normal string types everywhere? I have changed a lot of places to NewFromUtf8Literal which saves some typing but Internalized strings should behave differently, I think.

Makefile
pkit: Done!

plv8_func.cc
pkit: Streamlined with `NewFromUtf8Literal`, still ugly, but less...

pkit:
Added `Reallocate` for array buffers.
Added callback for unhandled promises
Added real handling for interrupts (SIGINT, SIGTERM), not clear how to test these though...

pkit:
Found out that there is a problem here when some exceptions can lead to segfault in postgress.
Unfortunately repro is pretty complicated. Will update when I find why it happens...

JerrySievert:
that's pretty much why I stopped at 8.7 in my branch - I was able to segfault on a regular basis just by running `make installcheck` on macOS (not Big Sur, so not the `EINTR` issue).

pkit:
`make installcheck` works fine each time
it's only nested exception that goes south for some reason

P.S. it builds under 20 min here, will bisect :)

JerrySievert:
it worked for me on linux - hence my calling it out for macos.  very weird.

JerrySievert:
also worked for me when I would run the tests one by one in `psql` as opposed to running the install check.

pkit:
> also worked for me when I would run the tests one by one in `psql` as opposed to running the install check.

this shit I remember, was fixed by re-arranging some handlers to always do `if (someValue.toLocal(&localValue)) { ... }`

pkit:
Essentially any toLocalChecked() can segfault in 8.8 ......

pkit:
Okay, found it. It's because of the newest addition of Promise rejection callback which gets called even for synchronous code sometimes it seems.
Will investigate what's going on there. But it looks like you cannot handle exception in the callback.
You need to store it somewhere and can only handle it when your `->Call()` returns. Which is fishy but that's how it works in all the implementations I could find.

pkit:
Bumped, should work :)

JerrySievert:
ok, I'm going to get GitHub actions integrated later today, so I can feel better than "all checks have failed" and running it locally on my Mac and vm's.

pkit:
Hold your horses, maybe I screwed up. 
I'm cherry-picking changes from another branch and may have missed something.
Although from git diff it looks like everything that's needed is there, but will retest only that.

JerrySievert:
well, me getting GitHub actions configured won't necessarily mean the PR is complete and accepted, it just makes me feel better :)

JerrySievert:
ok, I think I've decided what I want to do.  I'm going to release 3.0, with the current 3.0alpha v8 version or updated slightly.

then going to create a branch for 3.1, merge this in, and start work on true native jsonb access.  that should round out a good chunk of 3.1.

give me a couple of days and I'll get it all set up.

JerrySievert:
I discovered while doing a ton of jsonb aggregate work with plv8, that the jsonb conversion might be better if it never had to convert.  99% of time was moving between jsonb and v8 JSON, and back to jsonb.

pkit:
@JerrySievert that's interesting, because it indeed seems to make some unneeded conversions. Will check it out.

Meanwhile I have a pretty much working solution for 8.8, i.e. it doesn't fail in production (pretty small scale though).
But I have a lot of changes to the core. And main new feature: custom contexts. Where you can have different `globalThis` for different stored procedures. But it kinda undermines `start_proc` though, so I have an admin only `boot_proc` for that, but probably need a more elegant solution for upstream here.

pkit:
closing in favor of #441

PR #434, opened 2021-04-22
Use a specific version of depot_tools

thogrefe-foreflight:
The current version of `plv8` clones `depot_tools` but does not specify a version or commit, so `plv8` ends up grabbing the most recent version of `depot_tools`. This provides a way for unexpected changes to enter `plv8` via `depot_tools`. This PR locks the version of `depot_tools` to a specific commit that we know works, keeping the behavior of `plv8` the same.


JerrySievert:
closing - I'm abandoning gclient and depot_tools.

PR #469, opened 2021-12-27
build: Move CCFLAGS to Makefile.shared

jtojnar:
These defines are required even for distros that build v8 separately so they use `make -f Makefile.shared`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,20 +134,6 @@ v8_rebuild: v8_config v8
 
 include Makefile.shared
 
-ifdef EXECUTION_TIMEOUT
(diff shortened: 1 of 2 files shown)


JerrySievert:
I'm afraid that this won't be merged.  3.1 will include a new makefile and build system to work around v8's horrible requirements and builds.

JerrySievert:
no longer applies, going to go ahead and close this.  feel free to open another PR against the current build system if you find it inadequate.

jtojnar:
Yeah, it is still inadequate for my purpose. Opened https://github.com/plv8/plv8/pull/505.

PR #483, opened 2022-04-06
Update plv8.md

CaliViking:
Added notes regarding problems I ran into on Ubuntu 20.04 LTS - I am not sure that I have placed them in the right part of the document.

This file is out of date and needs some love. The apt-get install postgresql-{your-postgresql-version-here}-plv8 instructions no longer seems to be valid.

Would it make sense to structure the document so that it had a section per operating system? The dependencies and instructions are very different in each OS.

### DIFF
--- a/doc/plv8.md
+++ b/doc/plv8.md
@@ -34,18 +34,25 @@ PL/v8 is tested with:
 - g++: version 4.8.2
 - clang++
 
+Here is an example of dependencies that will likely be used when building on Ubuntu 20.04 LTS (change postgresql-server-dev-12 to your version of postgres)
+```bash
+   sudo apt install g++ python build-essential autoconf automake gdb git libffi-dev zlib1g-dev libssl-dev apt-file libtinfo5 postgresql-server-dev-12 ninja-build
+```
+
+To install the V8 pre-requisites, see https://github.com/v8/v8 . Be aware that it can take a long time to build V8 from source.
+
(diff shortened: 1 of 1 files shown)


doc/plv8.md
```diff
@@ -34,18 +34,25 @@ PL/v8 is tested with:
...
+To install the V8 pre-requisites, see https://github.com/v8/v8 . Be aware that it can take a long time to build V8 from source.
+
 Also all tools that PostgreSQL and V8 require to be built are required if you
 are building those from source.
 
 ## Installing PL/v8
 
 ### Build from source:
-Determine the [PL/v8 release](https://github.com/plv8/plv8/releases) you want to download and use it's version and path below.
+Determine the [PL/v8 release](https://github.com/plv8/plv8/releases or https://github.com/plv8/plv8/tags) you want to download and use it's version and path below.
```
JerrySievert, line 50: that's 2 url's and an `or` in that url path.  can you make them separate please?

JerrySievert:
thanks for the PR.

I've added a quick note for a change, but we also don't use the `master` branch, but the current version instead.

also note that the full documentation (which is also built for plv8.github.io) is under `docs`, and while these still are used, they are only used as single file for `PGXN`, which can only handle a single markdown file.

I'm wondering if it might be better to minimize what is in `doc/plv8.md` and have it mostly point to plv8.github.io instead.

PR #484, opened 2022-04-12
Properly escape shell PATH in Linux Makefile

malakhite:
The Linux Makefile currently uses the `$$PATH` syntax to use access the shell's `$PATH`. This fails when the `PATH` contains special characters (`()&` etc.). Quoting the reference and using `${PATH}` to access the variable fixes this problem.


JerrySievert:
awesome, thanks for you contribution!

Issue #459, opened 2021-09-28
Cannot stop process if executing with plv8.execute


JamesInform:
Using: plv8 2.3.15

If I run the following statement
`do
$$
	plv8.execute("select pg_sleep(180)");
$$ language plv8;
`

I cannot stop the statement before the 180 seconds are over.

Wether 

`select pg_terminate_backend(pid) `

nor 

`select pg_cancel_backend(pid)`

is working.

What can I do?

JerrySievert:
which operating system for the database, and which version of postgres?

JamesInform:
I can reproduce it unter
Mac PG13, PG14
Ubuntu 20.04 PG13, PG14

alessbelli:
Hi @JerrySievert - this was affecting our setup after we upgraded RDS versions, and this got down to a chase for the regression.
At this point this seems to work fine on any plv8 version until v2.3.3:
- you can either ctrl+c or pg_cancel_backend(pid) to kill long-running execution of inner SQL commands
- you can set a statement timeout that will get respected (#439)
- pure js infinite loops cannot be terminated  (issue #187)

(tested using [waldo2188/postgres-plv8](https://github.com/waldo2188/docker-postgres-plv8/blob/master/Dockerfile) on docker, which runs pg10 + plv8 2.3.3)

starting v2.3.4:
- pure js infinite loops seem to be terminatable with pg_cancel_backend / ctrl + c
- plv8 that runs long-running SQL commands can't get terminated
- plv8 execution doesn't respect statement timeouts.

(tested using [capitalch/xpostgres-plv8](https://github.com/capitalch/xpostgres-plv8/blob/master/Dockerfile) on docker, which runs pg10 + plv8 2.3.4 (the dockerfile doesn't match what's actually installed!)

The tests I ran were pretty simple:
- cancellability of inner SQL commands:
    - `DO $$ plv8.execute("SELECT pg_sleep(60)") $$ LANGUAGE plv8;`
- statement timeouts:
    - `BEGIN TRANSACTION; set local statement_timeout to 1000; DO $$ plv8.execute('select pg_sleep(2);') $$ LANGUAGE plv8; COMMIT;`
- infinite js loops
    - `DO $$ a = 1; while (true) {a+=1;}; $$ LANGUAGE plv8;`

[...]

PR #494, opened 2022-08-09
Add cancellations to clear termination interrupts.

sfingram:
The v8 isolate method `IsExecutionTerminating` only checks for if the termination exception has been thrown, not if the interrupt has been thrown.  If the interrupt has been thrown at the end of a cancellation, it will generate the exception on a possibly-unrelated query at a later time.  We can use the `interrupted` flag in the `current_context` to track whether we've been interrupted and then call `CancelTerminateExecution` which resets the interrupt, preventing the potential for an unrelated termination exception being thrown the next time we use the isolate.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -770,8 +770,11 @@ DoCall(Local<Context> ctx, Handle<Function> fn, Handle<Object> receiver,
 	Isolate 	   *isolate = ctx->GetIsolate();
 	TryCatch		try_catch(isolate);
 
-	if (isolate->IsExecutionTerminating()) {
+	if (isolate->IsExecutionTerminating() || current_context->interrupted) {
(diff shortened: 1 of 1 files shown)


sfingram:
This is discussed [at length here](https://github.com/plv8/plv8/issues/459#issuecomment-1203141997).

This PR should close this issue, @JerrySievert .

JerrySievert:
thanks @sfingram - I took a quick look at it, but not enough in-depth.  I'll try to look again tomorrow, or at the latest this weekend.

JerrySievert:
this looks great, thank you!

I'm going to merge now, but not do a new release immediately.  I have a couple of other small fixups I want to include in the next couple of days.

PR #498, opened 2022-08-31
Update Dockerfile

hellower:
strip command not found
need bintuils package


JerrySievert:
huh, interesting.  I wonder if that's a difference between the arm64 vs x86 - I'm assuming you used x86?

hellower:
yes.  x86

JerrySievert:
awesome, thanks!

PR #504, opened 2022-10-12
Initialize unhandled_promises in plv8_context

kelvich:
Otherwise, the `unhandled_promises` vector can be initialized with garbage. A better solution would be something like placement new, but since we are using Postgres memory allocator, it is tricky. With zeroing out during the initialization, we depend on the exact implementation of the vector. As far as I know, zeroing out will set `len` = `cap` = 0 and will work correctly afterward.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -1903,7 +1903,7 @@ GetPlv8Context() {
 	}
 	if (!my_context)
 	{
-		my_context = (plv8_context *) MemoryContextAlloc(TopMemoryContext,
+		my_context = (plv8_context *) MemoryContextAllocZero(TopMemoryContext,
 														 sizeof(plv8_context));
 		my_context->is_dead = false;
 		my_context->interrupted = false;


kelvich:
Ok, rewrote it to the placement new instead

plv8.cc
```diff
@@ -1903,8 +1903,7 @@ GetPlv8Context() {
 	}
 	if (!my_context)
 	{
-		my_context = (plv8_context *) MemoryContextAlloc(TopMemoryContext,
-														 sizeof(plv8_context));
+		my_context = new(MemoryContextAlloc(TopMemoryContext, sizeof(plv8_context))) plv8_context();
```
JerrySievert: we should not be using `new` here - I know it's c++ but we still need to work within the TopMemoryContext - what are you trying to accomplish?  there's likely a better way that is better for Postgres.

kelvich: Well, we need somehow to call the default constructor for the vector, otherwise, vector itself (namely internal `len/capacity` or `_begin/_end`) contains some random data, and we cannot use it.

It is placement new, so it just calls the constructors for `plv8_context()` fields and places result in memory that was allocated via `MemoryContextAlloc(TopMemoryContext, ...)`. So I think it is okay in terms of postgres memory context tracking. Things that vector will allocate on the heap will escape Postgres memory contexts, but that will happen irregardless of the way we initialize it.

Another option is just to use `MemoryContextAllocZero`, which is not guaranteed to properly initialize vector, but I guess will work with all of the existing implementations of a vector.

JerrySievert: > Another option is just to use MemoryContextAllocZero, which is not guaranteed to properly initialize vector, but I guess will work with all of the existing implementations of a vector.

it should - I'd prefer the `MemoryContextAllocZero`, but I am still unable to reproduce the issue.

> Things that vector will allocate on the heap will escape Postgres memory contexts, but that will happen irregardless of the way we initialize it.

yeah, it's unfortunate.  we also lose track of a small amount of other things that get allocated, but thankfully not enough to be a substantial memory leak (one thing v8 is pretty good about is memory management, as long as you provide it a _decent_ allocator).

kelvich: > but I am still unable to reproduce the issue.

I think you can put some deterministic garbage in memory referenced by this pointer to force a crash, e.g.
```c
// let's fill that memory with deterministic garbage
for (int i = 0; i < sizeof(plv8_context); i++)
    *((char *)my_context + i) = i;
```
Standalone example of the same problem: https://gist.github.com/kelvich/e8b63ddcf4af3ce63afb26d92083afac

> it should - I'd prefer the MemoryContextAllocZero, but I am still unable to reproduce the issue.

Looking at https://github.com/llvm-mirror/libcxx/blob/master/include/vector#L342-L343 it probably will, but anyway, I think placement new is a better way -- AFAIK it won't allocate anything and will put initialized structure in buffer returned by `MemoryContextAlloc`. That way proper constructor will be called for std::vector.

Having said I'm not against `MemoryContextAllocZero` and can update this patch to use it (first commit used it, actually).

JerrySievert: I have a horrible 4-letter word that unfortunately applies to this project that makes me wary of a placement new: msvc

I think I'd prefer the `MemoryContextAllocZero`, just based on that alone.

so I can try to reproduce more easily, which version of clang++ did you compile against, or was it g++?

kelvich: > I have a horrible 4-letter word that unfortunately applies to this project that makes me wary of a placement new: msvc

Gotcha :)

> so I can try to reproduce more easily, which version of clang++ did you compile against, or was it g++?

`gcc version 10.2.1 20210110 (Debian 10.2.1-6)`, one that goes with debian:bullseye, https://gist.github.com/kelvich/5ced631a4210991147f1dda254c6b5c1

but if you put garbage in `my_context` right after allocation, I'm pretty sure it will crash. I observed a few times that systems with low memory pressure tend to return zeroed pages from allocations -- I think that is probably your case.

JerrySievert:
> gcc version 10.2.1 20210110 (Debian 10.2.1-6), one that goes with debian:bullseye

official plv8 dockerfile uses clang/clang++ - which definitely behaves differently than gcc/g++ - don't even get me started on stuff that passed `purify` with flying colors 30 years ago that crash just by looking at them :) (I'd actually be curious if you can reproduce with a modification of that dockerfile for postgis).

regardless, if you want to go back to the `MemoryContextAllocZero` I'd probably merge it just on principle, but I think I still want to reproduce before merging.

thanks for the PR, it is appreciated, and as soon as I can do a full test I'll merge and put out another release.

I have a new server coming tonight - can't promise I'll be able to get to it all before this weekend, but compiling v8 will be a good burn-in for the server 😂

kelvich:
Ok, pushed `MemoryContextAllocZero` version.

JerrySievert:
good news, I've managed to replicate - and now that I've replicated I have the perfect test case.

giving it a go now.

JerrySievert:
works fantastically, thank you very much for your PR!

PR #505, opened 2022-10-13
build: Allow using V8 from system

jtojnar:
Building V8 is slow and pointless on systems that already provide a recent version like some Linux distros.
With this change, you can build against system V8 with the following:

    make USE_SYSTEM_V8=1 SHLIB_LINK=-lv8 V8_OUTDIR=/usr/lib

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ OBJS = $(SRCS:.cc=.o)
 MODULE_big = plv8-$(PLV8_VERSION)
 EXTENSION = plv8
 PLV8_DATA = plv8.control plv8--$(PLV8_VERSION).sql $(wildcard upgrade/*.sql)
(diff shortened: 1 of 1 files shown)


JerrySievert:
this is not a PR that will be accepted.

while it may work for you in particular, for this exact version of plv8 against that exact version of v8 for that exact version of the distro that you are using at this exact second, it will not work elsewhere.

the v8 project specifically is not meant to be compiled as a shared library:

* the ABI changes between minor versions, making it incompatible
* the API changes between minor versions, making it incompatible
* any options that are different during compilation (because v8 is designed to be configurable for the specific embedded target) will cause random crashes

so, after a very large support burden trying to support shared libraries on even a single distribution, let alone on 2+, the decision was made to disable linking against a system copy of libv8.

that decision will not be reversed.

Issue #466, opened 2021-12-09
Upgrading the extension


jtojnar:
Last time I interacted with my application I was running plv8 2.3.13. Now I need to use it again but in the meanwhile, the distro I use switched to 2.3.15, causing most of the app queries to fail with:

```
ERROR:  could not access file "$libdir/plv8-2.3.13": No such file or directory
```

I also tried to update the extension as specified in the docs but it failed with the same error:

```
=# ALTER EXTENSION plv8 UPDATE;
ERROR:  could not access file "$libdir/plv8-2.3.13": No such file or directory
```

I cannot just drop the extension because the rest of the database depends on it:

```
=# DROP EXTENSION plv8;
ERROR:  cannot drop extension plv8 because other objects depend on it
```

I also tried updating the plv8 package to 3.0 but that went even worse since update path does not seem to be defined:

https://github.com/plv8/plv8/blob/v3.0.0/generate_upgrade.sh#L6

```
=# ALTER EXTENSION plv8 UPDATE;
ERROR:  extension "plv8" has no update path from version "2.3.13" to version "3.0.0"
```

And if I patch the package similarly to https://github.com/plv8/plv8/pull/465, it again fails with the familiar error:

```
=# ALTER EXTENSION plv8 UPDATE;
ERROR:  could not access file "$libdir/plv8-2.3.13": No such file or directory
```

I suppose I can install the old `plv8-2.3.13.so` back but is this really required for upgrade to work?

Or I could run `UPDATE pg_proc SET probin = '$libdir/plv8-3.0.0' WHERE proname LIKE 'plv8%';` but I feel rather queasy about tweaking Postgres’s internal state.

I use PostgreSQL 11.14 on NixOS.

patrickmichalina:
Same issue here. What is the correct procedure?

JerrySievert:
@patrickmichalina there should be upgrade paths for everything in 3.1.6, is there something you see missing?

patrickmichalina:
The only thing I can think of is my custom docker image is not including the previous versions?

```bash
# plv8
COPY --from=builder /usr/lib/postgresql/${PG_MAJOR}/lib/plv8* /usr/lib/postgresql/${PG_MAJOR}/lib/
COPY --from=builder /usr/lib/postgresql/${PG_MAJOR}/lib/bitcode/plv8-${PLV8_VERSION}.index.bc /usr/lib/postgresql/${PG_MAJOR}/lib/bitcode/
[...]

PR #465, opened 2021-12-09
generate_upgrade.sh: Add older versions

jtojnar:

PR #506, opened 2022-10-13
generate_upgrade.sh: Add older versions

jtojnar:
Otherwise the upgrade from 2.3.13 to 3.0.0 would fail with the following error:

    =# ALTER EXTENSION plv8 UPDATE;
    ERROR: extension "plv8" has no update path from version "2.3.13" to version "3.0.0"

There are still some manual migration steps necessary, see https://github.com/plv8/plv8/issues/466.
But at least this patch fixes this specific one.

----

Resurrecting https://github.com/plv8/plv8/pull/465.


JerrySievert:
quick question:

are these giving you benefit, or are you submitting for completionism?

if the former, I'd love a use case.

thanks!

jtojnar:
Sorry, I have made the commit message more descriptive. When I tried to upgrade to 3.0.0, it failed with the following error, until I applied this patch:

    ERROR:  extension "plv8" has no update path from version "2.3.13" to version "3.0.0"

Even with this, manual migration steps are still necessary, see https://github.com/plv8/plv8/issues/466, but at least it is a progress.

JerrySievert:
we should now be completely up to date, I'm going to close this.

if you feel it is in error, that one or more were missed, feel free to reopen.

PR #514, opened 2023-01-19
Fix function argument name to avoid clash with SQL/JSON reserved word

donsergioq:
In case of SQL/JSON implementation (like in Postgres Pro) the word 'json' will be reserved word. So, it shouldn't be used as an argument name in function to avoid test failure.


JerrySievert:
lgtm, thanks much for the PR!

PR #515, opened 2023-01-19
Use the v8 default allocator; necessary in sandbox

maparent:
I am building plv8 against more recent versions of v8. The most recent has the option of using a sandbox mode. In that mode, it is necessary to use the plv8 default allocator. Since it has no constructor, it is easiest to define the plv8 allocator as a facade. This should also work for earlier v8 versions.


JerrySievert:
plv8 is explicitly tied to v8 versions and does not support other versions by design.  sorry, I cannot accept this until plv8 moves forward with another major release (which are tied to v8 versions).

JerrySievert:
as with #516, I've implemented these changes as part of another plv8 update that is not yet pushed - so going to close this.  thanks for the PR though!

PR #516, opened 2023-01-19
Flags can be frozen by v8::Initialize. Set flags first

maparent:
Again needed by later v8 versions.


JerrySievert:
see note on #515.

JerrySievert:
thanks, going to close since I implemented this on a newer non-committed version of plv8.

after some other changes are done (external to plv8) I'll push the changes to a new branch.

PR #517, opened 2023-01-24
Update a list of older plv8 versions

donsergioq:


JerrySievert:
awesome, thank you!

would you do me a quick favor (since you already have the PR open) and fill in the blanks between 2.3.10 and 2.3.15 in the main upgrade script?  it would be much appreciated.  might as well fix them all while we're here.

JerrySievert:
thanks much!

PR #519, opened 2023-02-19
Add configs patching into Makefiles

donsergioq:


JerrySievert:
do you have any existing patches that should be added?

JerrySievert:
this looks good, thanks @donsergioq!

PR #524, opened 2023-03-12
Improve Docker build

abrobston:
The current Docker build is a great start.  However, it is slower and less flexible than it could be, and it does not complete successfully.

This PR:
- Updates the Dockerfile to support more container build systems (like `buildah`)
- Allows setting the Postgres version, plv8 version, and plv8 branch via `--build-arg` or equivalent
- Clones only the necessary revisions of git repositories in both Dockerfile and Makefile.docker
- Patches `install-sysroot.py` to eliminate a build error due to inadequate `tar` flags

The patch comes in two files because later versions of the Chromium build already use the patch from the first file


JerrySievert:
awesome, thanks for the PR!  I have a couple of comments that I'll open up for discussion before I begin any testing of it.

Makefiles/Makefile.docker
```diff
@@ -8,41 +8,38 @@ ifeq ($(ARCH),x86_64)
...
 GN_ARGS = is_component_build=false v8_use_external_startup_data=false v8_enable_i18n_support=false is_clang=false use_gold=false use_custom_libcxx=false treat_warnings_as_errors=false v8_monolithic=true
 
 
 build/gn/out/gn:
 	@echo "Setting up build directory"
 	@mkdir -p build
 	@echo "Downloading gn source"
-	@git clone -q https://gn.googlesource.com/gn.git build/gn
-	@cd build/gn && git checkout -q eea3906f0e2a8d3622080127d2005ff214d51383
+	@$(call clone_all_to_hash,build/gn,https://gn.googlesource.com/gn.git,eea3906f0e2a8d3622080127d2005ff214d51383)
```
JerrySievert: I'm not a big fan of this.  I understand the DRY aspect of it, but it really makes it a lot harder to quickly understand exactly what steps are being done in the download and build process.

if this could be unrolled, I'd be much happier.

Makefiles/Makefile.docker
```diff
@@ -8,41 +8,38 @@ ifeq ($(ARCH),x86_64)
...
-	@cd build/v8/third_party/googletest/src && git checkout -q 47f819c3ca54fb602f432904443e00a0a1fe2f42
-	@git clone -q https://chromium.googlesource.com/chromium/deps/icu.git build/v8/third_party/icu
-	@cd build/v8/third_party/icu && git checkout -q 75e34bcccea0be165c31fdb278b3712c516c5876
-	@git clone -q https://chromium.googlesource.com/chromium/src/third_party/jinja2.git build/v8/third_party/jinja2
-	@cd build/v8/third_party/jinja2 && git checkout -q ee69aa00ee8536f61db6a451f3858745cf587de6
-	@git clone -q https://chromium.googlesource.com/chromium/src/third_party/markupsafe.git build/v8/third_party/markupsafe
-	@cd build/v8/third_party/markupsafe && git checkout -q 1b882ef6372b58bfd55a3285f37ed801be9137cd
-	@git clone -q https://chromium.googlesource.com/chromium/src/third_party/zlib.git build/v8/third_party/zlib
-	@cd build/v8/third_party/zlib && git checkout -q 563140dd9c24f84bf40919196e9e7666d351cc0d
+	@$(call clone_hash,build/v8/base/trace_event/common,https://chromium.googlesource.com/chromium/src/base/trace_event/common.git,3da1e2fcf66acd5c7194497b4285ac163f32e239)
```
JerrySievert: same comment as the `clone_all_to_hash` above.

Makefiles/Makefile.docker
```diff
@@ -8,41 +8,38 @@ ifeq ($(ARCH),x86_64)
...
+	@$(call clone_hash,build/v8/build,https://chromium.googlesource.com/chromium/src/build.git,bbf7f0ed65548c4df862d2a2748e3a9b908a3217)
+	@$(call clone_hash,build/v8/third_party/googletest/src,https://chromium.googlesource.com/external/github.com/google/googletest.git,47f819c3ca54fb602f432904443e00a0a1fe2f42)
+	@$(call clone_hash,build/v8/third_party/icu,https://chromium.googlesource.com/chromium/deps/icu.git,75e34bcccea0be165c31fdb278b3712c516c5876)
+	@$(call clone_hash,build/v8/third_party/jinja2,https://chromium.googlesource.com/chromium/src/third_party/jinja2.git,ee69aa00ee8536f61db6a451f3858745cf587de6)
+	@$(call clone_hash,build/v8/third_party/markupsafe,https://chromium.googlesource.com/chromium/src/third_party/markupsafe.git,1b882ef6372b58bfd55a3285f37ed801be9137cd)
+	@$(call clone_hash,build/v8/third_party/zlib,https://chromium.googlesource.com/chromium/src/third_party/zlib.git,563140dd9c24f84bf40919196e9e7666d351cc0d)
 	@cp Makefiles/gclient_args.gni build/v8/build/config/gclient_args.gni
 	@echo "Setting up sysroot"
+	@cd build/v8/build && (git apply ../../../patches/sysroot/*.patch || true)
```
JerrySievert, line 65: when patching files, please add an `@echo Patching [...]`, it helps make it much clearer when attempting to debug build issues (which are unfortunately common with v8).

Makefiles/Makefile.docker
abrobston: Consider it unrolled. 🙂

Makefiles/Makefile.docker
abrobston: I unrolled these calls to multiple lines each for clarity, even if starting with `cd ` every time gets redundant.

Makefiles/Makefile.docker
abrobston: Done.

Makefiles/Makefile.docker
JerrySievert: awesome, thanks!

JerrySievert:
I am just finishing https://github.com/plv8/plv8/pull/523 which finally adds back CI after the fall of travis, I will add a docker build step to include the testing of this.

JerrySievert:
thank you very much!  I had been waiting on #523 but haven't had the time to finish it up.  merged, but did not release as part of the v3.1.6 release.

PR #545, opened 2023-06-30
Wrap `this` keyword with backticks

imageck:


JerrySievert:
awesome, thanks for noticing and the PR!

PR #409, opened 2020-10-02
fix upgrade-related woes with GUC redefinitions

pkit:
fixes #333 
have no idea how to test the upgrade path though

PR #552, opened 2023-07-31
fix upgrade-related woes with GUC redefinitions

pkit:
Forward-port of #409


JerrySievert:
Awesome! Thanks. I should be able to merge tomorrow and I’ll handle the 16 changes. Much appreciated.

PR #558, opened 2023-09-22
 fixed bug with scalar JSONB arguments wrongly interpreted as array

deem0n:
Hi, we are switching from plv8 2.x to 3.x and have found a bug caused by JSONB_DIRECT_CONVERSION.

How to reproduce:

```
mi=> CREATE OR REPLACE FUNCTION jsonb_cat(data jsonb) RETURNS jsonb LANGUAGE plv8 AS
$$  return data; $$;

mi=> SELECT jsonb_cat(to_jsonb('"asd"'::TEXT));
 jsonb_cat
-----------
 ["asd"]
(1 row)

mi=> SELECT jsonb_cat(to_jsonb(1::INT));
 jsonb_cat
-----------
 [1]
(1 row)

mi=> SELECT jsonb_cat(to_jsonb(false::BOOL));
 jsonb_cat
-----------
 [false]
(1 row)
```

We expect scalar JSON value returned, not array. Obvious fix is provided. Thanks for plv8, it is a joy!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ endif
 
 DATA = $(PLV8_DATA)
 DATA_built = plv8.sql
-REGRESS = init-extension plv8 plv8-errors inline json startup_pre startup varparam json_conv \
+REGRESS = init-extension plv8 plv8-errors scalar_args inline json startup_pre startup varparam json_conv \
 		  jsonb_conv window guc es6 arraybuffer composites currentresource startup_perms bytea find_function_perms \
 		  memory_limits reset show array_spread regression procedure
 
--- a/expected/scalar_args.out
+++ b/expected/scalar_args.out
@@ -1,0 +1,59 @@
+CREATE OR REPLACE FUNCTION jsonb_scalar_cat(data jsonb) RETURNS jsonb LANGUAGE plv8 AS
+$$ plv8.elog(NOTICE, 'arg = ' + JSON.stringify(data));  return data; $$;
+SELECT jsonb_scalar_cat(to_jsonb('asd'::TEXT));
+NOTICE:  arg = "asd"
+ jsonb_scalar_cat 
+------------------
+ "asd"
+(1 row)
+
+SELECT jsonb_scalar_cat(to_jsonb(19450509::INT));
+NOTICE:  arg = 19450509
+ jsonb_scalar_cat 
+------------------
+ 19450509
+(1 row)
+
+SELECT jsonb_scalar_cat(to_jsonb(false::BOOL));
+NOTICE:  arg = false
+ jsonb_scalar_cat 
+------------------
+ false
+(1 row)
+
+SELECT jsonb_scalar_cat('{"key":[null, true, false, 19450509, "string"]}'::JSONB);
+NOTICE:  arg = {"key":[null,true,false,19450509,"string"]}
+                 jsonb_scalar_cat                 
+--------------------------------------------------
+ {"key": [null, true, false, 19450509, "string"]}
+(1 row)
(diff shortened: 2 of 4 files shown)


JerrySievert:
thanks for the PR!

can you please add a couple of tests (feel free to create a new `sql/` and `expected/` file) to help validate this in the future?

deem0n:
Hi, yep tests are on the way. But I found that return value is also converted to array, not only input args.

```sql
CREATE OR REPLACE FUNCTION jsonb_cat(data jsonb) RETURNS jsonb LANGUAGE plv8 AS
$$  return 'str'; $$;
CREATE FUNCTION
contrib_regression=# select jsonb_cat('[]'::JSONB);
 jsonb_cat
-----------
 ["str"]
(1 row)
```

Expecting just scalar "str"...

Can you point me to the file/function  where we can fix this? Any hint?

deem0n:
Hi, I found [place](https://github.com/plv8/plv8/blob/f23425b5115203d7b339123d5088bf82bfff51cc/plv8_type.cc#L543) where we can fix return values. Last thing is to handle null JSONB values. I will post PR soon.

deem0n:
While I've been fixing the input/output scalar bug, introduced with JSONB_DIRECT_CONVERSION I found 2 gotchas:

1. In versions without JSONB_DIRECT_CONVERSION (2.x & 3.x) it is not possible to correctly return scalar string from the v8 engine!

```sql
CREATE OR REPLACE FUNCTION jsonb_glitch1()
 RETURNS jsonb
 LANGUAGE plv8
AS $$ return "glitch"; $$;

select jsonb_glitch1();
ERROR:  invalid input syntax for type json
DETAIL:  Token "glitch" is invalid.
CONTEXT:  JSON data, line 1: glitch


CREATE OR REPLACE FUNCTION jsonb_glitch2()
 RETURNS jsonb
 LANGUAGE plv8
AS $$ return '"glitch"'; $$;

select jsonb_glitch2();
 jsonb_glitch2
---------------
 "glitch"
(1 row)
```

With extra quotes you still can return scalar strings, but that looks ugly. I'm surprised it was not discovered/fixed for a long time, but hopefully everybody passing arrays and objects around and nobody cares about scalars ;-)

Anyway version 3 with JSONB_DIRECT_CONVERSION has fixed it, so `jsonb_glitch1()` works and returns string as expected. `jsonb_glitch2()` returns `"\"glitch\""` and that is correct.

2. JSON null handling is interesting. `null` is one of the supported scalar values in JSON, so it might be useful to return it as JSON scalar null, not SQL NULL. We have `undefined` in JavaScript and that is not supported by JSON as scalar value. So we can use `undefined` to return SQL NULL.

Version 3 (with fixes from this PR) can accept both nulls as input scalars:

```
contrib_regression=# SELECT jsonb_scalar_cat('null'::JSONB);
NOTICE:  arg = null
 jsonb_scalar_cat
------------------

(1 row)

contrib_regression=# SELECT jsonb_scalar_cat(NULL::JSONB);
NOTICE:  arg = null
 jsonb_scalar_cat
------------------

(1 row)
```
Input SQL NULL might be interpreted as `undefined` though.

Possible change:

JavaScript return    | SQL receive | Proposed 
-------------------|-------------|-----------
  return null;            |  NULL           |  'null'::JSONB or 'null'::JSON
  return undefined; |  NULL           |  NULL

Proposed changes is one line fix and I can do another PR if desired.

JerrySievert:
thanks much for the contribution - sorry it took so long to merge!  I will also look at back porting this to the 3.1 branch.

PR #561, opened 2023-11-04
Bytea parameter crashes plv8 with out of memory error

andreas:
We're using plv8 for operations on binary data represented as `bytea` columns in Postgres. We've started seeing plv8 crashing Postgres with the error message `Out of memory error` (likely coming from [here](https://github.com/plv8/plv8/blob/f23425b5115203d7b339123d5088bf82bfff51cc/plv8.cc#L213)). In the Postgres logs, we see the following:

```
<--- Last few GCs --->
[29773:0x3fa7de0]    55701 ms: Mark-sweep (reduce) 0.8 (2.7) -> 0.8 (1.9) MB, 1.6 / 0.1 ms  (average mu = 0.011, current mu = 0.009) external memory pressure GC in old space requested
[29773:0x3fa7de0]    55702 ms: Mark-sweep (reduce) 0.8 (1.9) -> 0.8 (1.9) MB, 1.0 / 0.1 ms  (average mu = 0.011, current mu = 0.011) external memory pressure GC in old space requested
<--- JS stacktrace --->
```

We've managed to narrow it down to being associated with `bytea` parameters. Below is the smallest repro we've been able to construct that somewhat mirrors our use case:

```sql
create or replace function noop(data bytea) returns void as $$
$$ language plv8;

do $$
  for(let i = 0; i < 2000; i++) {
    plv8.execute(`
      select noop(repeat('0', 1024*1024)::bytea) from generate_series(0, 100)
    `);
  }
$$ language plv8;
```
The test passes a 1MB `bytea` value to a dummy plv8 function for 100 rows, and then repeats the process up to 2,000 times. 
You can see this test case failing as part of the Github Actions workflow [here](https://github.com/andreas/plv8/actions/runs/6754701490/job/18362450361).

In practice, we process on the order of ~500 rows totalling ~16MB of bytea values for a single query. Over the course of a few minutes, queries of that shape might run many times over, which ultimately leads to the crash.

The repro can be simplified to the following, but departs further from our actual use case:

```sql
select count(noop(repeat('0', 1024*1024)::bytea)) from generate_series(0, 200*1000);
```

### DIFF
--- a/plv8_allocator.h
+++ b/plv8_allocator.h
@@ -10,8 +10,8 @@ class ArrayAllocator : public v8::ArrayBuffer::Allocator {
 private:
 	size_t heap_limit;
 	size_t heap_size;
-	size_t next_size;
-	size_t allocated;
+	std::atomic<size_t> next_size;
+	std::atomic<size_t> allocated;
 	v8::ArrayBuffer::Allocator* allocator;
 
 	bool check(size_t length);


JerrySievert:
I'm struggling to understand what this PR is attempting to do?  plv8 is not _crashing_, but returning an error that it is beyond its memory limits and cannot allocate more memory because of that.

have you checked the output of `show plv8.memory_limit`?   you're not disposing of the memory used from the output of your `plv8.execute()`, and are likely running up against a specifically set memory limit.

not disposing of results means that the gc cannot do cleanup.

andreas:
> I'm struggling to understand what this PR is attempting to do? plv8 is not crashing, but returning an error that it is beyond its memory limits and cannot allocate more memory because of that.

I should probably have opened an issue rather than a PR. I forked the repo so I could use `plv8`s Github Actions workflow to demonstrate the issue in a reproducible environment.

I recognise it looks like plv8 is just returning an error for a single query, but in practice we see (1) all existing connections to the Postgres instance being terminated with `Connection terminated unexpectedly`, (2) new connection attempts fail temporarily with `the database system is not yet accepting connections`.

> have you checked the output of show plv8.memory_limit? 

Our `plv8.memory_limit` is currently set to the default 256:

```
=> show plv8.memory_limit;
 plv8.memory_limit
-------------------
 256
(1 row)
```

We're looking into adjusting `plv8.memory_limit`, but it's unclear whether it would solve the problem or just delay how often we would experience crashes.

> you're not disposing of the memory used from the output of your plv8.execute(), and are likely running up against a specifically set memory limit.
>
> not disposing of results means that the gc cannot do cleanup.

I intentionally tried to create a very small, self-contained repro, but perhaps I golfed it down to a point where it's no longer valid (though it's not clear to me how the results of `plv8.execute()` should disposed to allow cleanup).

I'm quite confident our actual use case _does_ allow cleanup. The repro below using `node-postgres` mirrors our actual use case better:

```js
import pg from 'pg';
const client = new pg.Client();
await client.connect();

while (true) {
  await client.query("select count(noop(repeat('0', 1024*1024)::bytea)) from generate_series(0, 100)");
}
```

Is that a valid example, which should not cause an out of memory error (but does)?

JerrySievert:
v8 does not clean up ArrayBuffers, you can reproduce the same behavior in node.js and in chrome console with:

```
for (let i = 0; i < 10000; i++) {
    a = new Uint8Array(1024*1024);
}
```

the memory is allocated to the `arrayBuffers`, and not released.  you might be able to force a release of this by using `delete` on the incoming parameter.

I might be able to force deletion of the outgoing `ArrayBuffer`, but that will not help you much.  `arrayBuffers` do not appear to be cleaned up in v8 via garbage collection.

this is only one of the memory management issues of v8, and likely not something that can be addressed by plv8.

you are encouraged to look at https://github.com/plv8/pljs which aims to work around the large number of issues with v8.

andreas:
> v8 does not clean up ArrayBuffers, you can reproduce the same behavior in node.js and in chrome console with: ...

I'm not able to reproduce an OOM error in Node or Chrome with the provided snippet. Can you rule it this being specific to plv8? It seems like a pretty fundamental limitation if v8 does not clean up ArrayBuffers.

> you might be able to force a release of this by using `delete` on the incoming parameter.

I tried adding `delete`, but could not observe a difference in behaviour: it still triggers an OOM error 😕

> you are encouraged to look at https://github.com/plv8/pljs which aims to work around the large number of issues with v8.

Unfortunately pljs is not an option as we use on AWS RDS Postgres (which does not support pljs) and we depend heavily on WASM (which is listed as out of scope for pljs) 😕

JerrySievert:
your chrome and node.js maximum memory is set _much_ higher than the 256mb of limit that you have plv8 set to.  that is why you are not seeing the oom error directly.  you can look at the memory inspector in chrome and look at `process` in node.js to see the current memory usage and that it is not being garbage collected.

JerrySievert:
try executing
```
     return {
       next:() => this
     }
  };
  [...({})];
  ```
  
  in chrome or node.js to see how bad memory management in v8 is.

**Comment by andreas** [comment]
Do you have any ideas for how to work around this problem? We didn't experience this problem when we rolled out our solution based on plv8, and by now it's a pretty critical dependency with extensive usage. I can't be certain there haven't been other influencing factors, but I don't think we experienced this while being on plv8 v2.3.15.

**Comment by JerrySievert** [comment]
plv8 2.3 was a _much_ older version of v8 (6.4) compared to modern plv8, which uses 11.5.  many things have changed, including the bulk of v8 itself.  (6.4 is not compatible with 11.5, these compatibility changes are why plv8 versions are tied to v8 versions).

plv8 (and thus v8) engines are tied directly to the postgres process, and since postgres uses a forking model, disconnecting your bouncer/pool to a single instance will destroy that instance, and thus the engine itself.

you should be able to set a maximum usage for your pool/bouncer to disconnect from postgres, which will destroy that forked process, and a new connection will end up with a new process and thus a new instance of v8.

unfortunately, the only way to reliably get binary data in javascript is with an `ArrayBuffer`, which means there are no other real options for getting `bytea` data into javascript.

**Comment by andreas** [comment]
Setting a max usage has also been the only workaround we've identified ourselves – currently we've set a very low time-to-live for each connection so v8 is frequently reset. We haven't been able to proactively detect the issue by looking at `plv8.memory_usage
()`. The graph below shows the reported memory numbers leading up to a OOM error:

plv8 memory usage

Are you aware of any method to detect it in advance?

Thanks for all your help by the way, we really appreciate it 🙇 🙏

**Comment by JerrySievert** [comment]
other than looking at memory usage, I cannot think of any other way to predict unless you are able to understand how many of these you are running and do the math from there.

you can also raise your maximum memory for v8 (`plv8.memory_limit`) which can be raised to 3096mb, but that will have rippling effects throughout your database, so if you do that you need to make sure you have enough memory overhead available in the host.  rds/aurora are fairly opaque to me, so I do not know how buffers and other memory tunings compare to stock postgres.

**Comment by andreas** [comment]
I've spent some more time looking into this, and I have potentially found a bug. By adding debug logging, I noticed that `allocated` would have a value like `18446744073706405888` in `Check` right before an OOM error here:

https://github.com/plv8/plv8/blob/f23425b5115203d7b339123d5088bf82bfff51cc/plv8_allocator.cc#L23

Some more debug logging revealed, that the subtraction below in `Free` could cause `allocated` (and `next_size`) to wrap around:

https://github.com/plv8/plv8/blob/f23425b5115203d7b339123d5088bf82bfff51cc/plv8_allocator.cc#L58-L59

This leads the array allocator to believe it's exceeding the permitted allocated bytes and results in a OOM error.

To avoid the wrap around, I've replaced the subtraction in `Free` with the following (commit 7e1b855):

```js
	allocated = allocated > length ? allocated - length : 0;
	next_size = next_size > length ? next_size - length : RECHECK_INCREMENT;
```

It's unclear to me how this situation can arise, i.e. why is more memory freed than allocated 🤔 

With this change, I no longer see OOM errors for the use case mentioned before:

```js
import pg from 'pg';
const client = new pg.Client();
await client.connect();

while (true) {
  await client.query("select count(noop(repeat('0', 1024*1024)::bytea)) from generate_series(0, 100)");
}
```

JerrySievert:
thank you for the fantastic search and PR!

but, before I merge the PR, I'd rather try to track down the root cause, rather than patch a broken behavior.

andreas:
> thank you for the fantastic search and PR!
> 
> but, before I merge the PR, I'd rather try to track down the root cause, rather than patch a broken behavior.

I agree it would be better to find the root cause. I'll try to do some more investigation. Let me know if you have any leads 🙂

andreas:
I've spent some more time on this, and I now believe the issue is due to a race condition. The source code comment for `v8::ArrayBuffer::Allocator` says the following ([source](https://github.com/bnoordhuis/v8-cmake/blob/90d6634d820af06f14c642a9ef3a7bfa7ab4667a/v8/include/v8-array-buffer.h#L134C6-L134C78)):

> A thread-safe allocator that V8 uses to allocate |ArrayBuffer|'s memory.

However, the plv8 allocator is not thread-safe. By using `std::atomic<size_t>` for `allocated` and `next_size` (commit 78e6ad0), I no longer see them wrapping around. This also means I no longer see OOM errors for the previously mentioned use case 🥳

andreas:
@JerrySievert, what are your thoughts on merging this PR?

JerrySievert:
I have something else I want to get done before doing a merge, so that a release can be made.

andreas:
Any news on this? Let me know if I can help.

JerrySievert:
thanks for the awesome work tracking this down, and for being patient.  I have together what should be good release of 3.2.1 today, and will work on back porting this patch to the 3.1 branch and making a release there as well.

PR #562, opened 2023-11-08
Fix code blocks in BUILTINS.md

tmr08c:
There were additional backticks in the documentation for `find_function` which led to incorrect formatting:

Screenshot 2023-11-08 at 11 12 28 AM

This change removes the extraneous backticks, fixing the formatting.


JerrySievert:
good find, and thanks for the contribution!

PR #570, opened 2024-01-24
Add workflow to release on PGXN for a semver tag

### LINKED ISSUE #569: PGXN Build Fails
Looks like the `Makefile` sets up some `git submodule` stuff. You can use [git-archive-all](/Kentzo/git-archive-all) to includes submodules in the archive to submit to PGXN, which should eliminate the need to call `git submodule`.

theory:
Just set `PGXN_USERNAME` and `PGXN_PASSWORD` in the project secrets and it should just do it on the next semver tag.

Also use the `GIT_ARCHIVE_CMD` variable so that the bundle includes all submodules, teach PGXN not to index them, and tweak the `Makefile` to avoid calling `git` if the submodule is already included. Fixes #569.

Check out [this build](https://github.com/theory/plv8/actions/runs/7646998561/job/20836995424) to see it in action (without actually publishing to PGXN). The "Bundle Release" step shows it including the submodule, and the "Release on PGXN" step runs `ls -lah plv8-*.zip` to show the final bundle at 47MB.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+.gitattributes export-ignore
+.github export-ignore
+.gitignore export-ignore
+.gitmodules export-ignore
--- a/.github/workflows/pgxn-release.yml
+++ b/.github/workflows/pgxn-release.yml
@@ -1,0 +1,21 @@
+name: 🚀 Release on PGXN
+on:
+  push:
+    # Release on semantic version tag.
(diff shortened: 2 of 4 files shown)


JerrySievert:
awesome!  thank you!  merging now, but waiting for my new pgxn account request to go through to set up the secrets.

PR #590, opened 2024-10-24
Upgrade to cpp20 for v8 11.6 support

alexeyfadeev:
v8-cmake was upgraded to cpp20. cpp20 is needed to build v8 11.6 with v8-cmake

Tested with PostgreSQL 17 on Debian and Aplpine.


JerrySievert:
thanks for the PR, but plv8 (like node.js) does not do major v8 upgrades on minor plv8 versions.

PR #591, opened 2024-10-29
Fix possible overflow in Window Functions

eeshan9815:
The issue is that the variable `size` is of the type `size_t`, which is an unsigned integer. 

Passing a negative value leads to an integer overflow, and the requested size is in the order of 2^64 bytes, which is over 10^10 GB, leading to a heap overflow. 

Calling the window function will now lead to an error (ERROR:  allocation size cannot be negative) instead of crashing.


JerrySievert:
thanks much!  letting the tests run now

PR #593, opened 2024-11-04
Match plv8 external array types by their element type, in addition to their name.

eeshan9815:
The issue is that plv8_fill_type() matches types only by their names but not their shapes or element types.

This PR fixes that by checking the element type and only allowing predefined types (int2, int4, int8, float4, float8) for v8's typed arrays.


JerrySievert:
do you happen to have any tests that can show incorrect behavior?

thanks!

PR #595, opened 2025-04-09
Fix Build on PG 18

jakob:
Attributes are no longer directly accessible in PG 18. This PR changes the code so that the appropriate macro is used instead.

I'm not sure which is your current development branch, so I wasn't sure which branch to base the PR on.


JerrySievert:
thanks for the PR - normally I wait until commitfest has completed to add support for the next version, as there are accelerated changes during that period that could be backward incompatible, but I will assess these changes in the meantime.

and yep, definitely the correct branch!

JerrySievert:
oh, yes, this is quite innocuous, thanks much for the contribution!

PR #602, opened 2025-10-17
Update upgrade generation script to allow upgrading from 3.1.9 and 3.…

esiaero:
…1.10

Hi, this seems like an oversight that there is no scripts made for 3.1.9 and 3.1.10, so I thought I'd send a patch for it.


JerrySievert:
hi!  thanks for the PR, but I have to ask: do you have a use case for this?  plv8 3.1 reach end of life of support seven months ago, and I'm hesitant to make any changes or a release.

esiaero:
Just to clarify, this is not about supporting 3.1.x versions, its more to make it possible to migrate those 3.1.x versions to 3.2.4. Right now there's no upgrade scripts for it on 3.2.x versions.

JerrySievert:
I appreciate the PR, but 3.1 is EOL, and since its release there has not been an issue with this (in 99.9% of cases, 3.1.9 and 3.1.10 included), it's a no-op (especially since the windows support was tenuous as best), I'm going to close without acting on it.

PR #605, opened 2025-11-24
Added ability to write language handlers in Javascript

postsql:
The "language handler is currently a transpiler from function source to javascript

a sample usage - a language which transpiles any source to "return 42"

language handler which just transpiles any source to "return 42;"

```sql
CREATE OR REPLACE FUNCTION pseudo_handler()
RETURNS language_handler
LANGUAGE plv8
AS $$
   plv8.elog(NOTICE, "source is [", arguments[0], "]");
   return "return 42";
 $$;
-- CREATE FUNCTION
```
create a language using this handler
```sql
CREATE OR REPLACE TRUSTED LANGUAGE always42 HANDLER pseudo_handler;
-- CREATE LANGUAGE
```
create a function in this newly created language
```sql
CREATE OR REPLACE FUNCTION test_always42() RETURNS int LANGUAGE always42 AS $$ return o; $$;
-- CREATE FUNCTION
```
run the function
```sql
SELECT * FROM test_always42();
NOTICE:  source is [  return o;  ]
 test_always42
---------------
            42
(1 row)
```

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -12,22 +12,24 @@
 #endif
 
 #include "libplatform/libplatform.h"
+#include "plv8_allocator.h"
 
-#include <new>
 
 extern "C" {
-#if PG_VERSION_NUM >= 90300
 #include "access/htup_details.h"
-#endif
 #include "access/xact.h"
+#include "catalog/pg_language.h"
 #include "catalog/pg_proc.h"
+#include "catalog/pg_database.h"
 #include "catalog/pg_type.h"
 #include "commands/trigger.h"
+#include "common/hashfn.h"
 #include "executor/spi.h"
 #include "funcapi.h"
 #include "miscadmin.h"
 #include "utils/builtins.h"
 #include "utils/guc.h"
+#include "utils/guc_tables.h"
 #include "utils/memutils.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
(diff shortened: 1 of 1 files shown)


postsql:
## The things that would be also nice to have

- **`VALIDATOR` support** - in case the validator function is written in pl/v8 pass it argument list with types and return type so it can validate that the defined transpiled language can check these
  - I plan to write pl_jsonschema which allows writing JSON Schema validators by wrapping [AJV](https://ajv.js.org/) and the validator should allow only  a single `JSON` argument and a `boolean` return value
- **Documentation**
- **let the handler return a function**:  pl/v8 should recognize when the handler function returns a callable (as AJV does by default) and then use the returned object directly (with currently implemented strict source-to-javascript transpiler architecture I have to extract the function source and have pl/v8 compile it again)

## Longer term: WASM support

- **raw arguments/return values**:  I already can do WASM support, but it would be more efficient to skip the default argument and return value conversion to Javascript and let WAS handle PostgreSQL data formats directly
  - this would work better if we already compiled core type support tow WASM, like we do for LLM for JIT support
- **support in Rust and other to generate pl/v8 flavour of WASM** - toolkits already package their results for node.js and browsers, they could also support pl/v8

JerrySievert:
I've added a `r4.0` branch that this can be actively developed in, also added a roadmap.

PR #607, opened 2026-04-14
Fix crash: Use cached isolate for V8 scopes in plv8_call_handler

### LINKED ISSUE #606: Abend/crash when executing plv8 function in a different security context
**Describe the bug**
Execution of plv8 routines from within a SECURITY DEFINER routine, where said routine owner != execution user may cause a crash within plv8

The issue is that common_pl_call_handler (plv8_call_handler in plv8 v3.2+) makes incorrect assumptions about the context it obtains from GetPlv8Context() with respect to the context that may already be associated with the function being executed. That is, consider the following scenario with the preconditions
 
* plv8 function F
* pgplsql procedure P, cataloged as SECURITY DEFINER privUser, which executes function F statements run as normalUser

```
connect to database as normalUser
select F()
call P()
```

results in the following series of events
 
    [select F()]
    common_pl_call_handler (function F) -> GetPlv8Context()
    GetPlv8Context finds the context for the current user (ie: normalUser)
    create IsolateScope/HandleScope based on curcontext->isolate
    CompileFunction
    CreateExecEnv() based on current context, assign to fcinfo->flinfo->fn_extra
    CallFunction (function F) using the exec env created
    [call P()]
    [select f() inside of P]
    common_pl_call_handler() -> GetPlv8Context()
    GetPlv8Context find the context for the current user. Which is actually privUser, because the user context has been changed by the execution of the SEC DEF routine
    create IsolateScope/HandleScope based on curcontext->isolate
    CompileFunction/CreateExecEnv is NOT called - we pull the information fcinfo out of the sqlfunc cache, so fn_extra is already set
    CallFunction (function F) using the exec env pulled from fn_extra

At this point, we're broken. The context that CallFunction is operating with is NOT the same as the context pulled at the top of common_pl_call_handler, and therefore we're using a different isolate from the one we setup the IsolateScope/HandleScope with on entry to the function.

The lack of a HandleScope is made very apparent the first time we try to create any handle whatsoever, when we hard abort in the v8 engine code:

```
  // Make sure there's at least one scope on the stack and that the
  // top of the scope stack isn't a barrier.
  if (!Utils::ApiCheck(current->level != current->sealed_level,
                       "v8::HandleScope::CreateHandle()",
                       "Cannot create a handle without a HandleScope")) {
    return nullptr;
  } 
```

**To Reproduce**

```
create or replace function jrt_plv8() returns int as
$$
  var i = 1;
  plv8.elog(NOTICE, "hi there");
  return i;
$$ language plv8;

create or replace function jrt_plv8wrapper() returns void as
$$
DECLARE
  tmp int := 0;
begin
  tmp := jrt_plv8();
END;
$$ language plpgsql;

create or replace procedure jrt_plpgsql() as
$$
begin
  raise warning 'about to do somethign';
  perform jrt_plv8wrapper();
  raise warning 'did somethiong';
end;
$$ language plpgsql security definer;

set role josh;
begin;
select jrt_plv8wrapper();
call jrt_plpgsql();
select jrt_plv8wrapper();
end;

```

**Expected behavior**
No v8 engine assertion.

**Version Information:**
 - PLV8 Version: 3.1.10
 - Postgres Version 17.7
 - Operating System and Version Amazon Linux 2

**Additional context**
The naive solution appears to be to force CallFunction (and CallTrigger/CallSRFFunction) to setup IsolateScopes/HandleScopes based on the isolate in the execution environment, rather than from the current context.

eg:

```
diff --git a/plv8.cc b/plv8.cc
index c1ce883..406cb3a 100644
--- a/plv8.cc
+++ b/plv8.cc
@@ -513,8 +513,11 @@ common_pl_call_handler(PG_FUNCTION_ARGS, Dialect dialect) throw()
                        fcinfo->flinfo->fn_extra = proc;
                }
 
+               {
                plv8_proc *proc = (plv8_proc *) fcinfo->flinfo->fn_extra;
                plv8_proc_cache *cache = proc->cache;
+               Isolate::Scope  scope(proc->xenv->isolate);
+               HandleScope     handle_scope(proc->xenv->isolate);
 
                if (is_trigger)
                        return CallTrigger(fcinfo, proc->xenv);
@@ -524,6 +527,7 @@ common_pl_call_handler(PG_FUNCTION_ARGS, Dialect dialect) throw()
                else
                        return CallFunction(fcinfo, proc->xenv,
                                                cache->nargs, proc->argtypes, &proc->rettype);
+               }
        }
        catch (js_error& e)     { e.rethrow(); }
        catch (pg_error& e)     { e.rethrow(); }
```

Note: this MAY not strictly require a SEC DEF routine to intermediate. Anything that triggers a role change after a function being executed in the current session is likely to also trigger this problem. ie: SET ROLE (or moral equivalents)

leson238:
Fixes #606

## Problem

When a plv8 function gets cached and then called through a `SECURITY DEFINER` routine (or after `SET ROLE`), the effective user changes. `GetPlv8Context()` resolves by effective user, so it returns a different isolate than the one the function was compiled under. The `Isolate::Scope`/`HandleScope` at the top of `plv8_call_handler` are set up with this new isolate, but the cached function still points to the old one, and this mismatch is causing server crashes.

## Fix

After cache retrieval, open an inner block scope with `Isolate::Scope` and `HandleScope` from `proc->xenv->isolate` before dispatching. This way V8 scopes always match the isolate the function was actually compiled under and added minimum runtime overhead cost.

## Reproduction (unfixed code)

Using the repro from #606:

```sql
CREATE FUNCTION jrt_plv8() RETURNS int AS $$
  var i = 1;
  plv8.elog(NOTICE, "hi there");
  return i;
$$ LANGUAGE plv8;

CREATE FUNCTION jrt_plv8wrapper() RETURNS void AS $$
DECLARE tmp int := 0;
BEGIN tmp := jrt_plv8(); END;
$$ LANGUAGE plpgsql;

CREATE PROCEDURE jrt_plpgsql() AS $$
BEGIN
  RAISE WARNING 'about to do something';
  PERFORM jrt_plv8wrapper();
  RAISE WARNING 'did something';
END;
$$ LANGUAGE plpgsql SECURITY DEFINER;

CREATE ROLE josh LOGIN;
-- grant execute on all three to josh

SET ROLE josh;
BEGIN;
SELECT jrt_plv8wrapper();   -- caches plv8 func under josh's isolate
CALL jrt_plpgsql();          -- SECURITY DEFINER switches to superuser context
SELECT jrt_plv8wrapper();
END;
```

Server crashes:
```
NOTICE:  hi there
 jrt_plv8wrapper
-----------------

(1 row)

WARNING:  about to do something
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
connection to server was lost
```

## After fix

Same repro, no crash:
```
NOTICE:  hi there
 jrt_plv8wrapper
-----------------

(1 row)

WARNING:  about to do something
NOTICE:  hi there
WARNING:  did something
CALL
NOTICE:  hi there
 jrt_plv8wrapper
-----------------

(1 row)

COMMIT
```

Full regression suite passes:
```
ok 1  - init-extension
ok 2  - plv8
...
ok 26 - procedure
ok 27 - security_definer
ok 28 - bigint
# All 28 tests passed.
```

## Changes

- `plv8.cc` — inner block scope with `Isolate::Scope`/`HandleScope` from `proc->xenv->isolate` around function dispatch
- `Makefile` — register `security_definer` in REGRESS
- `sql/security_definer.sql` + `expected/security_definer.out` — regression test

## Environment

- PLV8 3.2.4, PostgreSQL 17.7, macOS ARM64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ DATA = $(PLV8_DATA)
 DATA_built = plv8.sql
 REGRESS = init-extension plv8 plv8-errors scalar_args inline json startup_pre startup varparam json_conv \
 		  jsonb_conv window guc es6 arraybuffer composites currentresource startup_perms bytea find_function_perms \
-		  memory_limits reset show array_spread regression procedure
+		  memory_limits reset show array_spread regression procedure security_definer
 
 ifndef BIGINT_GRACEFUL
 	REGRESS += bigint
--- a/expected/security_definer.out
+++ b/expected/security_definer.out
@@ -1,0 +1,74 @@
+-- Regression test for https://github.com/plv8/plv8/issues/606
+-- Crash when a cached plv8 function is called from a different security context.
+CREATE FUNCTION sd_plv8_func() RETURNS int AS $$
+  var i = 1;
+  plv8.elog(NOTICE, "hi there");
+  return i;
+$$ LANGUAGE plv8;
+CREATE FUNCTION sd_plv8_wrapper() RETURNS void AS $$
+DECLARE
+  tmp int := 0;
+BEGIN
+  tmp := sd_plv8_func();
+END;
+$$ LANGUAGE plpgsql;
+CREATE PROCEDURE sd_secdef_proc() AS $$
+BEGIN
+  RAISE WARNING 'about to call plv8 via wrapper';
+  PERFORM sd_plv8_wrapper();
+  RAISE WARNING 'plv8 call succeeded';
+END;
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+CREATE ROLE sd_test_role LOGIN;
+GRANT USAGE ON SCHEMA public TO sd_test_role;
+GRANT EXECUTE ON FUNCTION sd_plv8_func() TO sd_test_role;
+GRANT EXECUTE ON FUNCTION sd_plv8_wrapper() TO sd_test_role;
+GRANT EXECUTE ON PROCEDURE sd_secdef_proc() TO sd_test_role;
+-- Cache the plv8 function under sd_test_role's isolate
+SET ROLE sd_test_role;
+SELECT sd_plv8_wrapper();
+NOTICE:  hi there
+ sd_plv8_wrapper 
+-----------------
+ 
+(1 row)
+
+-- The procedure is owned by the superuser, so GetPlv8Context() switches
+-- to a different isolate while fn_extra still holds sd_test_role's.
+-- On unfixed code this crashes the server.
+CALL sd_secdef_proc();
+WARNING:  about to call plv8 via wrapper
+NOTICE:  hi there
+WARNING:  plv8 call succeeded
+-- Preservation: plv8 still works after the cross-context call
+SELECT sd_plv8_wrapper();
+NOTICE:  hi there
+ sd_plv8_wrapper 
+-----------------
+ 
+(1 row)
+
+RESET ROLE;
+-- Reverse direction: superuser cache, then SET ROLE
+SELECT sd_plv8_func();
+NOTICE:  hi there
+ sd_plv8_func 
+--------------
+            1
+(1 row)
+
+SET ROLE sd_test_role;
+SELECT sd_plv8_func();
+NOTICE:  hi there
+ sd_plv8_func 
+--------------
+            1
+(1 row)
+
+RESET ROLE;
+-- Cleanup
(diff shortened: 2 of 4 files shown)


JerrySievert:
thanks for the PR.  this didn't end up fixing the whole issue.  I've merged a fix that takes care of this completely though.

Issue #608, opened 2026-04-21
pg19 support


sibowu-aws:
**Describe the bug**
when complie with pg19 for plv8 3.2.4 version

```
o plv8_type.o plv8_type.cc
plv8_type.cc: In function ‘JsonbValue* JsonbFromValue(JsonbParseState**, v8::Local<v8::Value>, JsonbIteratorToken)’:
plv8_type.cc:447:31: error: cannot convert ‘JsonbParseState**’ to ‘JsonbInState*’
  447 |         return pushJsonbValue(pstate, type, &val);
      |                               ^~~~~~
      |                               |
      |                               JsonbParseState**
In file included from plv8_type.cc:21:
build/postgres/include/server/utils/jsonb.h:434:42: note:   initializing argument 1 of ‘void pushJsonbValue(JsonbInState*, JsonbIteratorToken, JsonbValue*)’
  434 | extern void pushJsonbValue(JsonbInState *pstate,
      |                            ~~~~~~~~~~~~~~^~~~~~
plv8_type.cc: In function ‘JsonbValue* JsonbArrayFromArray(JsonbParseState**, v8::Local<v8::Object>)’:
plv8_type.cc:454:42: error: cannot convert ‘JsonbParseState**’ to ‘JsonbInState*’
  454 |         JsonbValue *val = pushJsonbValue(pstate, WJB_BEGIN_ARRAY, NULL);
      |                                          ^~~~~~
      |                                          |
      |                                          JsonbParseState**
build/postgres/include/server/utils/jsonb.h:434:42: note:   initializing argument 1 of ‘void pushJsonbValue(JsonbInState*, JsonbIteratorToken, JsonbValue*)’
  434 | extern void pushJsonbValue(JsonbInState *pstate,
      |                            ~~~~~~~~~~~~~~^~~~~~
plv8_type.cc:468:30: error: cannot convert ‘JsonbParseState**’ to ‘JsonbInState*’
  468 |         val = pushJsonbValue(pstate, WJB_END_ARRAY, NULL);
      |                              ^~~~~~
      |                              |
      |                              JsonbParseState**
build/postgres/include/server/utils/jsonb.h:434:42: note:   initializing argument 1 of ‘void pushJsonbValue(JsonbInState*, JsonbIteratorToken, JsonbValue*)’
  434 | extern void pushJsonbValue(JsonbInState *pstate,
      |                            ~~~~~~~~~~~~~~^~~~~~
[...]

PR #609, opened 2026-04-21
Add Plv8 support for pg19: add missing hsearch.h include and adapt to…

sibowu-aws:
In PG19, utils/hsearch.h is no longer transitively included, causing
undeclared HASHCTL/hash_create/hash_search symbols in plv8.cc.

Also, pushJsonbValue changed signature from
  JsonbValue* pushJsonbValue(JsonbParseState**, ...)
to
  void pushJsonbValue(JsonbInState*, ...)

Updated plv8_type.cc to use JsonbInState and retrieve results via
pstate.result instead of the return value

related to https://github.com/plv8/plv8/issues/608

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -33,6 +33,7 @@ extern "C" {
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 #include "utils/syscache.h"
+#include "utils/hsearch.h"
 
 #include <signal.h>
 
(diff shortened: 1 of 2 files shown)


JerrySievert:
thanks for the PR!

there are a couple of issues why it cannot be accepted as-is:

1. it breaks all other supported Postgres versions, please see the [postgres versioning support](https://github.com/plv8/plv8/blob/r3.2/docs/VERSIONING.md#postgres-version-support): we support the last 4 major releases of Postgres, which currently are 15, 16, 17, and 18
2. there are no tags or branches currently in the postgres source tree to add to ci/cd.  until there is a tag or a branch we cannot test it in ci/cd

I'll leave the PR open and you're welcome to update it to fix both of these issues, but it cannot be merged as-is.

PR #611, opened 2026-05-07
Use PG_MODULE_MAGIC_EXT in PostgreSQL 18 and later

jeltz:
The `PG_MODULE_MAGIC_EXT` macro was added in PostgreSQL 18 and makes it possible to see which version of the library is actually loaded using `pg_get_loaded_modules()`.

### DIFF
--- a/plv8.cc
+++ b/plv8.cc
@@ -44,7 +44,14 @@ extern "C" {
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


JerrySievert:
awesome, thanks for the PR!  I hadn't added that in, and it was overdue!

